PR #3, opened 2017-05-28
Add the support necessary to be able to selectively run rules

Page-:


AbstractSQLCompiler.coffee
```diff
@@ -120,9 +176,15 @@
 				ruleBody = _.find(rule, 0: 'Body')[1]
 				ruleSE = _.find(rule, 0: 'StructuredEnglish')[1]
 				ruleSQL = compileRule(ruleBody, engine)
+				try
+					referencedFields = AbstractSQLCompiler.getReferencedFields(ruleBody)
```
dfunckt: @Page- this line here raises:

```
Error fetching referenced fields ReferenceError: AbstractSQLCompiler is not defined
at Object.compileSchema (/usr/src/app/node_modules/@resin/abstract-sql-compiler/AbstractSQLCompiler.js:246:32)
```

I think you just need `referencedFields = getReferencedFields(ruleBody)` to fix this.

PR #9, opened 2017-09-18
Fix rule compilation that would not pass bindings along

dfunckt:
This warrants a major version bump.

### DIFF
--- a/AbstractSQLRules2SQL.ometajs
+++ b/AbstractSQLRules2SQL.ometajs
@@ -738,5 +738,5 @@ export ometa AbstractSQLRules2SQL {
 			-> {query: query, bindings: this.fieldOrderings}
 		|	UpsertQuery('\n')
 		|	AnyValue('\n'):value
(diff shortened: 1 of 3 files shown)


AbstractSQLRules2SQL.ometajs
```diff
@@ -738,5 +738,5 @@ export ometa AbstractSQLRules2SQL {
 			-> {query: query, bindings: this.fieldOrderings}
 		|	UpsertQuery('\n')
 		|	AnyValue('\n'):value
-			-> ('SELECT ' + value + ' AS "result";')
+			-> ({query: 'SELECT ' + value + ' AS "result";', bindings: this.fieldOrderings})
```
Page-: Can you add spaces inside the braces please, ie
```
({ query: 'SELECT ' + value + ' AS "result";', bindings: this.fieldOrderings })

dfunckt:
@Page- PR updated.

PR #11, opened 2017-09-22
Convert AbstractSQLCompiler to typescript and add typings for the OMeta parts

Page-:
Change-type: minor


AbstractSQLCompiler.ts
```diff
@@ -0,0 +1,347 @@
...
+}
+
+const dataTypeGen = (engine: Engines, { dataType, required, index, defaultValue }: AbstractSqlField): string => {
+	let requiredStr
+	if (required) {
+		requiredStr = ' NOT NULL'
+	} else {
+		requiredStr = ' NULL'
+	}
+	if (defaultValue != null) {
```
nazrhom: Can we use _.isNull for all the null checks? It looks more consistent with the rest

AbstractSQLCompiler.ts
```diff
@@ -0,0 +1,347 @@
...
+		defaultValue = ` DEFAULT ${defaultValue}`
+	} else {
+		defaultValue = ''
+	}
+	if (index == null) {
```
nazrhom: _.isNull

AbstractSQLCompiler.ts
Page-: `_.isNull` specifically checks `=== null`, whereas we want `== null`, and for this specific case I prefer having the check directly as there are no special cases (whereas for a lot of types you have a bunch of hoops to jump through to check)

PR #13, opened 2017-10-20
Migrations

nazrhom:
Added support for automatic migrations between two schemas.

Supported auto-migrations are: adding or removing terms, both fields and tables, renaming will only be performed when removing a synonym or synonymous form.


test/sbvr/test.coffee
```diff
@@ -28,6 +28,24 @@ module.exports = exports = (builtInVocab = false) ->
...
+				srcSchema = LF2AbstractSQLTranslator(srcLf, 'Process')
+
+				SBVRParser.reset()
+				dstLf = SBVRParser.matchAll(dst, 'Process')
+				dstSchema = LF2AbstractSQLTranslator(dstLf, 'Process')
+
+				migration = AbstractSQLCompiler.postgres.diffSchemas(srcSchema, dstSchema)
+				console.log('GOT MIGRATION', migration)
+			catch e
+				throw e
```
Page-: You might as well drop the `try/catch` altogether if you're just going to rethrow :P

test/sbvr/test.coffee
```diff
@@ -28,6 +28,24 @@ module.exports = exports = (builtInVocab = false) ->
...
+			for i in [0...Math.max(migration.length, expectation.length)]
+				expect(migration[i]).to.equal(expectation[i])
```
Page-: This should be able to be
```coffee
expect(migration).to.deep.equal(expectation)
```
and avoid the need for the loop

test/sbvr/test.coffee
```diff
@@ -28,6 +28,24 @@ module.exports = exports = (builtInVocab = false) ->
...
+				SBVRParser.reset()
+				srcLf = SBVRParser.matchAll(src, 'Process')
+				srcSchema = LF2AbstractSQLTranslator(srcLf, 'Process')
+
+				SBVRParser.reset()
+				dstLf = SBVRParser.matchAll(dst, 'Process')
+				dstSchema = LF2AbstractSQLTranslator(dstLf, 'Process')
+
+				migration = AbstractSQLCompiler.postgres.diffSchemas(srcSchema, dstSchema)
+				console.log('GOT MIGRATION', migration)
```
Page-: Debug info?

test/sbvr/migrations.coffee
```diff
@@ -0,0 +1,110 @@
...
+		''',
+		'''
+		Term: pilot
+		Term: plane
+
+		Fact type:  plane can be flown by pilot
+		''', [ '''
+			ALTER TABLE "pilot-can fly-plane"
+				RENAME TO "plane-can be flown by-pilot";
+				RENAME COLUMN "pilot" TO "can be flown by-pilot";
```
Page-: I'm pretty sure this is invalid and should be
```
ALTER TABLE "plane-can be flown by-pilot"
RENAME COLUMN "pilot" TO "can be flown by-pilot";

File: AbstractSQLCompiler.ts [hunk 5]
```diff
@@ -359,15 +367,164 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+		diffSchemas: _.partial(diffSchemas, _, _, engine, ifNotExists),
 		dataTypeValidate,
 		getReferencedFields,
 		getModifiedFields,
 	}
 }
 export const postgres = generateExport(Engines.postgres, true)
 export const mysql = generateExport(Engines.mysql, true)
-export const websql = generateExport(Engines.websql, false)
+export const websql = generateExport(Engines.websql, false)
```
Page-: Can you put the newline back please

AbstractSQLCompiler.ts
```diff
@@ -359,15 +367,164 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+	
+	const modFn: ReduceFn<Pair<AbstractSqlField>> = (acc, {src, dst}) => {
+		if (_.isEqual(src, dst)) {
+			return acc
+		}
+		if (_.isEqual(_.omit(src, ['fieldName', 'references']), _.omit(dst, ['fieldName', 'references']))) {
+			acc.push('RENAME COLUMN "' + src.fieldName + '" TO "' + dst.fieldName + '";\n\t')
+			return acc
+		}
+		throw Error('Can not migrate pre-existing field ' + src.fieldName + ' to ' + dst.fieldName)
```
Page-: You can use interpolation with
```typescript
throw Error(`Can not migrate pre-existing field ${src.fieldName} to ${dst.fieldName}`)

File: AbstractSQLCompiler.ts [hunk 7]
```diff
@@ -359,15 +367,164 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+	} else {
+		ifNotExistsStr = ''
+		ifExistsStr = ''
+	}
+	//  defaultTypes =
+	//  	'Short Text': '\'\''
+	//  	'Text': '\'\''
+	//  	'Integer': 0
+
+	let matchFn: MatchFn<AbstractSqlField> = (fieldArray, field) => {
```
Page-: I'm pretty sure this has an `any` return type, can you check and type it correctly

AbstractSQLCompiler.ts
```diff
@@ -359,15 +367,164 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+//  All these functions are treated as reducers, their first argument
+//  will be an accumulator of all the changes up to that point and
+//  they are expected to add their changes to it and return it.
+//  The second argument will be an array of [key, value] for
+//  insertions and deletions and [key, pair] for the modifications.
+//  The match function describes how to find the corresponding T inside a T[].
+//  The result of the match function will be either the element that should be matched to 
+//  the argument or undefined if no such element is found.
+const genDiff = <T>(insFn: ReduceFn<T>, delFn: ReduceFn<T>, modFn: ReduceFn<Pair<T>>, matchFn: MatchFn<T>, src: Array<T>, dst: Array<T>) => {
+	let split = genSplit(src, dst, matchFn)
```
Page-: I'd prefer const here

AbstractSQLCompiler.ts
```diff
@@ -283,6 +285,12 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
 		}
 	})
 
+	return { hasDependants: hasDependants, schemaDependencyMap: schemaDependencyMap }
+}
+
+const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNotExists: boolean): SqlModel => {
+	let { hasDependants, schemaDependencyMap } = mkSchemaDependencyMap(abstractSqlModel.tables, engine, ifNotExists)
```
Page-: I'd prefer `const` here

AbstractSQLCompiler.ts
```diff
@@ -45,37 +45,48 @@ export interface Relationship {
...
+export interface ResourceMap<T> {
+	[ resourceName: string ]: T
+}
+
 export interface ModifiedFields {
 	table: string
 	fields?: {}[]
 }
 
+type ReduceFn<T> = (accumulator: any, element: T) => any
```
Page-: I'd prefer this to be better typed, my guess is you can do that for your purposes as
```typescript
type ReduceFn<T> = (accumulator: T[], element: T) => T[]
```

AbstractSQLCompiler.ts
```diff
@@ -45,37 +45,48 @@ export interface Relationship {
...
 	rules: SqlRule[]
 	createSchema: string[]
 	dropSchema: string[]
 }
 
+export interface hasDependants {
+	[dependant: string]: true
+}
+
+export interface schemaDependencyMap {
```
Page-: Can you use `CamelCase` for these interfaces please, to distinguish them from vars

AbstractSQLCompiler.ts
```diff
@@ -359,15 +367,164 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+					let verb = dstTable.name.split('-').slice(1, -1).join(' ')
+					return !_.isUndefined(relations[verb])
+				})
+			} else return relations
+		} else {
+			return match
+		}
+	}
+
+	const insFn: ReduceFn<AbstractSqlTable> = (acc, table ) => {
```
Page-: You have a stray space in here

AbstractSQLCompiler.ts
```diff
@@ -359,15 +367,164 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+//  will be an accumulator of all the changes up to that point and
+//  they are expected to add their changes to it and return it.
+//  The second argument will be an array of [key, value] for
+//  insertions and deletions and [key, pair] for the modifications.
+//  The match function describes how to find the corresponding T inside a T[].
+//  The result of the match function will be either the element that should be matched to 
+//  the argument or undefined if no such element is found.
+const genDiff = <T>(insFn: ReduceFn<T>, delFn: ReduceFn<T>, modFn: ReduceFn<Pair<T>>, matchFn: MatchFn<T>, src: Array<T>, dst: Array<T>) => {
+	let split = genSplit(src, dst, matchFn)
+	return  _.flatten(_.reduce(split.ins, insFn,
```
Page-: I think all this code could be made much nice with either
```typescript
_.flatMap(split.mod, modFn)
.concat(_.flatMap(split.del, delFn))
.concat(_.flatMap(split.ins, insFn))
```
and then `return []` to include nothing, and `return [a, b, c]` to return multiple.  It avoids all the need for mutation and overall should end up much nicer imo.  The alternative if you only ever need to return 0 or 1 results (never > 1) is:
```typescript
_.map(split.mod, modFn)
.concat(_.map(split.del, delFn))
.concat(_.map(split.ins, insFn))
.reject(_.isNil)
```
and `return` for nothing, `return a` for something

test/sbvr/migrations.coffee
nazrhom: The first instruction is there to rename the the table representing the relation. I used https://www.postgresql.org/docs/9.1/static/sql-altertable.html for reference but I think that it is pretty consistently supported by different dbs.

AbstractSQLCompiler.ts
nazrhom: I am not sure I understand. Both supplied matchFn typecheck correctly for me and the type definition specifies the return must be `T | undefined`

test/sbvr/migrations.coffee
Page-: Yes, that reference should be valid in general, my point is that the generated SQL is invalid according to that reference.  Try running the create schema statements for the first model to set it up and then running the migration statements, I'm pretty sure they will fail

AbstractSQLCompiler.ts
Page-: Ah, you're right the return function will be taken from `MatchFn<AbstractSqlField>`, my mistake

test/sbvr/migrations.coffee
Page-: Oh, also the change was referring to just the line I commented on, for reference I would expect the full block to be:
```coffee
[ '''
	ALTER TABLE "pilot-can fly-plane"
	RENAME TO "plane-can be flown by-pilot";
	ALTER TABLE "plane-can be flown by-pilot"
	RENAME COLUMN "pilot" TO "can be flown by-pilot";
	ALTER TABLE "plane-can be flown by-pilot"
	RENAME COLUMN "can fly-plane" TO "plane";
''' ]
```
or
```coffee
[ '''
	ALTER TABLE "pilot-can fly-plane"
	RENAME TO "plane-can be flown by-pilot";
''', '''
	ALTER TABLE "plane-can be flown by-pilot"
	RENAME COLUMN "pilot" TO "can be flown by-pilot";
''', '''
	ALTER TABLE "plane-can be flown by-pilot"
	RENAME COLUMN "can fly-plane" TO "plane";
''' ]
```

test/sbvr/migrations.coffee
```diff
@@ -0,0 +1,111 @@
...
+		Term: pilot
+		Term: plane
+
+		Fact type:  plane can be flown by pilot
+		''', [ '''
+			ALTER TABLE "pilot-can fly-plane"
+				RENAME TO "plane-can be flown by-pilot";
+			ALTER TABLE "plane-can be flown by-pilot"
+				RENAME COLUMN "pilot" TO "can be flown by-pilot";
+				RENAME COLUMN "can fly-plane" TO "plane";\n\t
```
Page-: This still has the issue from https://github.com/resin-io-modules/abstract-sql-compiler/pull/13#discussion_r146668460
> Oh, also the change was referring to just the line I commented on, for reference I would expect the full block to be:
> ```coffee
> [ '''
> 	ALTER TABLE "pilot-can fly-plane"
> 	RENAME TO "plane-can be flown by-pilot";
> 	ALTER TABLE "plane-can be flown by-pilot"
> 	RENAME COLUMN "pilot" TO "can be flown by-pilot";
> 	ALTER TABLE "plane-can be flown by-pilot"
> 	RENAME COLUMN "can fly-plane" TO "plane";
> ''' ]
> ```
> or
> ```coffee
> [ '''
> 	ALTER TABLE "pilot-can fly-plane"
> 	RENAME TO "plane-can be flown by-pilot";
> ''', '''
> 	ALTER TABLE "plane-can be flown by-pilot"
> 	RENAME COLUMN "pilot" TO "can be flown by-pilot";
> ''', '''
> 	ALTER TABLE "plane-can be flown by-pilot"
> 	RENAME COLUMN "can fly-plane" TO "plane";
> ''' ]
> ```

nazrhom: Ohh I kept misreading, I thought that renames on columns where in the list of actions. I will change that everywhere so that each rename has its own alter table

AbstractSQLCompiler.ts
```diff
@@ -359,10 +367,152 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+		} 
+	}
+	, { ins: dst, del: src, mod: mod } )
+}
+//  genDiff expects src and dst arrays formatted as described in
+//  genSplit, and three functions to be called in the case of
+//  insertion, deletion or modification.
+//  All these functions are treated as reducers, their first argument
+//  will be an accumulator of all the changes up to that point and
+//  they are expected to add their changes to it and return it.
```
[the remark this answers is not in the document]
nazrhom: I changed them in the last commit, given the comments Page did. I removed every reference to reducers to make it clear. They are just mapping functions which are not expected to have side-effects, but just map to the corresponding sql statement.

AbstractSQLCompiler.ts
```diff
@@ -359,10 +367,152 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+	const delFn: ReduceFn<AbstractSqlField> = (field) => {
+		return 'DROP COLUMN ' + ifExistsStr + '"' + field.fieldName + '";\n\t'
+	}
+	
+	const modFn: ReduceFn<Pair<AbstractSqlField>> = ({src, dst}) => {
+		if (_.isEqual(src, dst)) {
+			return
+		}
+		if (_.isEqual(_.omit(src, ['fieldName', 'references']), _.omit(dst, ['fieldName', 'references']))) {
+			return 'RENAME COLUMN "' + src.fieldName + '" TO "' + dst.fieldName + '";\n\t'
```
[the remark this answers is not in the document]
nazrhom: Good point, I assumed postgre for now, I think that most of these are pretty standard across SQL dialects but Ill check each statement.

AbstractSQLCompiler.ts
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+	const srcSDM = mkSchemaDependencyMap(src.tables, engine, ifNotExists).schemaDependencyMap
+	const dstSDM = mkSchemaDependencyMap(dst.tables, engine, ifNotExists).schemaDependencyMap
+
+	const matchFn:MatchFn<AbstractSqlTable> = (tables, srcTable) => {
+		let match = _.find(tables, { name: srcTable.name })
+		if (_.isUndefined(match)) {
+			const relations = src.relationships[srcTable.name]
+			if (!_.isUndefined(relations)) {
+				return _.find(tables, (dstTable) => {
+					let verb = dstTable.name.split('-').slice(1, -1).join(' ')
```
Page-: Can you add a `Term form` migration test, I think it might interact badly with this and would like to see it tested

AbstractSQLCompiler.ts
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+					return !_.isUndefined(relations[verb])
```
Page-: This can be
```typescript
return relations[verb] != null

File: AbstractSQLCompiler.ts [hunk 19]
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+	}
+	
+	const modFn: ToSQLFn<Pair<AbstractSqlField>> = ({src, dst}) => {
+		if (_.isEqual(src, dst)) {
+			return
+		}
+		if (_.isEqual(_.omit(src, ['fieldName', 'references']), _.omit(dst, ['fieldName', 'references']))) {
+			return 'RENAME COLUMN "' + src.fieldName + '" TO "' + dst.fieldName + '";'
+		}
+		throw Error(`Can not migrate pre-existing field ${src.fieldName} to ${dst.fieldName}`)
```
Page-: Could you add the data type to the message, as you can get here with the field name being the same but data type different, and the current error message would be confusing in that case

AbstractSQLCompiler.ts
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+			if (_.isString(mappings[field.fieldName])) {
+				return _.find(fieldArray, { fieldName: mappings[field.fieldName] } )
+			}
+		} else {
+			return match
+		}
+	}
+
+	const insFn: ToSQLFn<AbstractSqlField> = (field) => {
+		// Columns are inserted as NULL for the time being, this is because we have no sensible way to automatically assign a default value
```
Page-: Some fields do have a default value, so in that case we should make it a NOT NULL column (assuming it's meant to be NOT NULL)

AbstractSQLCompiler.ts
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+		ifNotExistsStr = 'IF NOT EXISTS '
+		ifExistsStr = 'IF EXISTS ' 
+	} else {
+		ifNotExistsStr = ''
+		ifExistsStr = ''
+	}
+
+	const matchFn: MatchFn<AbstractSqlField> = (fieldArray, field) => {
+		let match = _.find(fieldArray, { fieldName: field.fieldName })
+		if (_.isUndefined(match)) {
```
Page-: Can you shortcircuit with
```typescript
if (match != null) {
	return match
}

File: AbstractSQLCompiler.ts [hunk 22]
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+	if (ifNotExists) {
+		ifNotExistsStr = 'IF NOT EXISTS '
+		ifExistsStr = 'IF EXISTS ' 
+	} else {
+		ifNotExistsStr = ''
+		ifExistsStr = ''
+	}
+
+	const matchFn: MatchFn<AbstractSqlField> = (fieldArray, field) => {
+		let match = _.find(fieldArray, { fieldName: field.fieldName })
```
Page-: This should be `const`

AbstractSQLCompiler.ts
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+//  The result of the match function will be either the element that should be matched to 
+//  the argument or undefined if no such element is found.
+const generateDiff = <T>(insFn: ToSQLFn<T>, delFn: ToSQLFn<T>, modFn: ToSQLFn<Pair<T>>, matchFn: MatchFn<T>, src: Array<T>, dst: Array<T>) => {
+	const split = generateSplit(src, dst, matchFn)
+
+	const diff = _.map(split.modified, modFn)
+	.concat(_.map(split.deleted, delFn))
+	.concat(_.map(split.inserted, insFn))
+	
+	return _.reject(diff, _.isUndefined)
```
Page-: I'd prefer `_.isNil` to catch both `null` and `undefined`

AbstractSQLCompiler.ts
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
 	}
 }
 
+const generateSplit = <T>(src:Array<T>, dst: Array<T>, matchFn: MatchFn<T>): Split<T> => {
+	let modified: Array<Pair<T>> = []
```
Page-: This can be `const`

AbstractSQLCompiler.ts
```diff
@@ -370,10 +378,150 @@ const compileSchema = (abstractSqlModel: AbstractSqlModel, engine: Engines, ifNo
...
+	return _.reduce(src, (acc, value) => {
+		const match = matchFn(acc.inserted, value)
+		if (_.isUndefined(match)) {
```
Page-: this can be `match == null`

AbstractSQLRules2SQL.js
```diff
@@ -44,7 +44,7 @@
     }, AbstractSQLRules2SQL = exports.AbstractSQLRules2SQL = OMeta._extend({
         NestedIndent: function(indent) {
             var $elf = this, _fromIdx = this.input.idx;
-            return indent + "	";
+            return indent + "\t";
```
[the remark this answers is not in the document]
nazrhom: Not sure about this, my `AbstractSQLRules2SQL.ometa` is the same as the one appearing on master, I think this is caused by some other commit that was merged without including the changes to `.js` files?

PR #14, opened 2017-11-29
Set up circleci auto-publishing

Page-:
Change-type: patch


Page-:
I fixed the build and locked the deps down, patch versions breaking the build is a pita

PR #21, opened 2018-04-24
Update tests to use odata-to-abstract-sql 1.2.0

Page-:
Change-type: patch


package.json
```diff
@@ -21,10 +21,9 @@
     "ometa-js": "^1.4.1"
   },
   "devDependencies": {
-    "@resin/abstract-sql-to-odata-schema": "^1.1.0",
```
Page-: This was removed because it's unused

PR #27, opened 2018-05-31
Use like for comparisons

Page-:
The big one here is the StartsWith change, switching to a like changes the way postgres handles things significantly to really speed up some queries (I saw ~60% improvement on the device env var translations which is one of our slowest, but it should also help with all the cases of the SDK querying with `startswith(uuid, 'aaa')`)


dfunckt:
> should also help with all the cases of the SDK querying with startswith(uuid, 'aaa')

Yeah, the Dashboard does this request every few seconds when in Device Summary page. So, many 👍

AbstractSQLOptimiser.ometajs
```diff
@@ -1,5 +1,22 @@
 var _ = require('lodash');
 
+var escapeForLike = function(str) {
+	return [
+		'Replace',
+		[	'Replace',
+			[	'Replace',
+			str,
```
dfunckt: I think you want that indented one more level.

Page-: Yeah you're right, I have the file recognised as JS because it's the best syntax highlighting I've found, but when moving code around the "invalid" ometa sections can confuse the auto indenter - fixed

Issue #34, opened 2018-07-26
Feature: Support for SQL statement `CASE/WHEN/ELSE/END`


richbayliss:
It would be good if the library supported the `CASE/WHEN/ELSE/END` query command.

PR #35, opened 2018-07-26
Add support for CASE/WHEN/ELSE/END

richbayliss:
There are some usecases which require this query command. It should be
added to allow support for generating from the abstract SQL.

Change-type: patch
Connects-to: #34
Signed-off-by: Rich Bayliss <rich@resin.io>


richbayliss:
Tests to follow...

AbstractSQLRules2SQL.ometajs
```diff
@@ -743,6 +744,23 @@ export ometa AbstractSQLRules2SQL {
...
+		]
+		-> (' CASE ' + whens.join(' ') + ' ELSE ' + val + ' END'),
+
+	When :indent =
+		[
+			'When'
+			BooleanValue(indent):matches
+			AnyValue(indent):resultValue
+		]
+		-> (' WHEN ' + matches + ' THEN ' + resultValue + ' '),
```
Page-: Do you know if we ever need brackets around the `matches` or `resultValue`?  Or is that automatically handled within them? (it might well be)

AbstractSQLRules2SQL.ometajs
```diff
@@ -743,6 +744,23 @@ export ometa AbstractSQLRules2SQL {
...
 		),
 
+	Case :indent =
+		[
+			'Case'
+			[	When(indent)+:whens
+			]
+			AnyValue(indent):val
+		]
+		-> (' CASE ' + whens.join(' ') + ' ELSE ' + val + ' END'),
```
Page-: We have the `indent`/`NestedIndent` stuff which might be nicer to use in order to print the sql as
```sql
SELECT CASE
		WHEN a THEN n
		WHEN b THEN o
		WHEN c THEN p
		WHEN d THEN q
		WHEN e THEN r
		WHEN f THEN s
		ELSE z
	END
FROM ...
```
instead of
```sql
SELECT CASE WHEN a THEN n WHEN b THEN o WHEN c THEN p WHEN d THEN q WHEN e THEN r WHEN f THEN s ELSE z END
FROM ...
```
in order to make debugging a little easier on us humans

AbstractSQLRules2SQL.ometajs
```diff
@@ -743,6 +744,23 @@ export ometa AbstractSQLRules2SQL {
 			)
 		),
 
+	Case :indent =
+		[
+			'Case'
+			[	When(indent)+:whens
```
Page-: I think this would be nicer as
```
'Case'
When(indent)+:whens
AnyValue(indent):val
```
so you can do
```
[	'Case',
	[ 'When', ... ]
	[ 'When', ... ]
	[ 'When', ... ]
	[ 'When', ... ]
	[ ... ]
]
```
instead, it just makes it a slightly nicer format imo

Page-: It's probably also worth making the `ELSE` explicit via `[ 'Else', ... ]`

AbstractSQLRules2SQL.ometajs
```diff
@@ -743,6 +744,23 @@ export ometa AbstractSQLRules2SQL {
...
+			]
+			AnyValue(indent):val
```
Page-: The else value is optional in postgres, so it's probably better to have this as `AnyValue(indent)?:val` and then something like `val ? ' ELSE ' + val : ''`

AbstractSQLRules2SQL.ometajs
richbayliss: I believe that because I set them as `BooleanValue` that the brackets would be handled as-per normal; not sure about the result value though. Do you have a scenario in mind?

AbstractSQLRules2SQL.ometajs
richbayliss: Good idea. I will try and look to other examples in the OMetaJS to see how to do this, but I am presuming that `nestedIndent` is just a variable for `\t` etc?

AbstractSQLRules2SQL.ometajs
richbayliss: Agreed on both parts.

AbstractSQLRules2SQL.ometajs
richbayliss: As above, it should be clear and optional.

AbstractSQLRules2SQL.ometajs
Page-: `NestedIndent(indent):nestedIndent` just adds an extra level of nesting (`\t`), and so will be something like `\n\t\t`.  So you should be able to do `.join(nestedIndent)` `+ nestedIndent +` etc in the various places to format it nicely - and passing the `nestedIndent` into the child functions to propogate the nesting level.
Overall the indentation is just a nice to have, but it is super nice for debugging

AbstractSQLRules2SQL.ometajs
Page-: I think it's fine actually and brackets will be automatically added on the sub-parts that need them and we don't need them here

Page-:
@richbayliss you should be able to do
```
[	'Else'
	AnyValue(indent):val
]?
```
(the `?` being the key) and then `val` will just be undefined if an else wasn't found

AbstractSQLRules2SQL.ometajs
```diff
@@ -743,6 +744,26 @@ export ometa AbstractSQLRules2SQL {
 			)
 		),
 
+	Case :indent =
+		[	'Case'
+			NestedIndent(indent):nestedIndent
+			[	When(indent)+:whens
```
Page-: This should be changed to just `When(indent)+:whens` to have the `['Case', ['When' ...` format instead of `['Case', [['When' ...`

Page-: You probably want to pass `nestedIndent` here instead of `indent` to get nice formatting in the cases where `When` uses it

AbstractSQLRules2SQL.ometajs
```diff
@@ -743,6 +744,26 @@ export ometa AbstractSQLRules2SQL {
...
+			]
+			[	'Else'
+				AnyValue(indent):val
```
Page-: You probably want to pass `nestedIndent` here instead of `indent` to get nice formatting in the cases where `AnyValue` uses it

AbstractSQLOptimiser.ometajs
```diff
@@ -206,6 +206,23 @@ ometa AbstractSQLValidator {
 		]
 		-> [['Where', boolStatement]],
 
+	Case =
+		[	'Case'
+			[ 	When+:whens
```
Page-: This should also become `When+:whens`

AbstractSQLRules2SQL.ometajs
richbayliss: I thought the same, but when I tried it I got errors. I will try again though.

PR #37, opened 2018-10-03
Update dependencies

Page-:
Change-type: patch
Signed-off-by: Pagan Gazzard <page@resin.io>


package.json
```diff
@@ -5,32 +5,31 @@
...
   },
   "repository": "https://github.com/resin-io-modules/abstract-sql-compiler.git",
   "author": "",
   "dependencies": {
     "@resin/sbvr-types": "^1.4.2",
-    "@types/bluebird": "3.5.20",
-    "@types/lodash": "4.14.109",
-    "@types/node": "8.10.19",
+    "@types/bluebird": "^3.5.24",
+    "@types/lodash": "^4.14.116",
```
thgreasi: I remember that at some point bluebird and lodah typings were getting patch releases that broke our builds.
I guess that since we are adding back the `^` they don't do that any more.

package.json
```diff
@@ -5,32 +5,31 @@
...
-    "@resin/odata-parser": "^0.3.5",
-    "@resin/odata-to-abstract-sql": "2.2.0",
+    "@resin/lf-to-abstract-sql": "2.0.7",
+    "@resin/odata-parser": "^0.3.6",
+    "@resin/odata-to-abstract-sql": "2.2.5",
     "@resin/sbvr-parser": "0.1.4",
-    "chai": "^4.1.2",
-    "coffee-script": "^1.12.7",
+    "chai": "^4.2.0",
+    "coffeescript": "^1.12.7",
```
thgreasi: We should probably also change https://github.com/resin-io-modules/abstract-sql-compiler/blob/59dad7b0f98e3fd5d5f29c1ce59a09032a4e3029/test/mocha.opts#L3

package.json
Page-: Yeah, I haven't noticed it recently at all, what I have definitely noticed is using fixed bluebird versions causing `type Bluebird is incompatible with type Bluebird` (or a version of that) due to having multiple versions of the bluebird types in a project that haven't been deduped

package.json
Page-: Done thanks

Page-:
The tests change are to match a bugfix in lf-to-abstract-sql where a join would be missed in the case of `FUNC(x/y)`.  The reason I switched to `was_trained_by__pilot/name` instead of `licence/name` was just because it was an easier way to fix it, but either would generate equivalent sql for the respective tables

PR #39, opened 2018-10-09
Update to sbvr-types 2.x

Page-:
`dataTypeValidate` now rejects with error objects instead of strings

Change-type: major
Signed-off-by: Pagan Gazzard <page@resin.io>


AbstractSQLCompiler.ts
```diff
@@ -97,28 +97,16 @@ export interface EngineInstance {
 	getModifiedFields: (abstractSqlQuery: AbstractSqlQuery) => undefined | ModifiedFields | Array<undefined | ModifiedFields>,
 }
 
-const validateTypes = _.mapValues(sbvrTypes, ({ validate }) => {
-	if (validate != null) {
-		return Promise.promisify(validate)
-	}
-})
+const validateTypes = _.mapValues(sbvrTypes, ({ validate }) => validate)
```
dfunckt: Can;t this be `_.mapValues(sbvrTypes, 'validate')`?

Page-: Sort of, but I'm pretty sure it messes with types (it also has worse performance but obviously that only matters in some very niche cases)

PR #44, opened 2018-11-13
Convert to typescript

Page-:
The two conversions combined work out to an ~91% performance improvement for the optimisation + compilation step in the tests, with the improvement being split roughly evenly between the two steps.

The output between the ometajs and typescript versions should be equivalent (all existing tests continue to pass in the same way, and resin-api tests also pass) and the code is fairly close to a 1-1 mapping.  The main changes are that the optimisation step now handles some cases where we don't require the first element to be a string denoting the node type and inserts the correct type - handling those cases lets the compilation step be simpler and paves the way to removing support for those cases altogether


AbstractSQLOptimiser.ts
```diff
@@ -0,0 +1,1217 @@
...
+	try {
+		return JSONValue(args);
+	} catch (e) {
+		CheckErr(e);
+	}
+	try {
+		return DurationValue(args);
+	} catch (e) {
+		CheckErr(e);
+	}
```
zvin: This could be written with a `for valueType of [TextValue, NumericValue, ...] {`

AbstractSQLRules2SQL.ts
```diff
@@ -0,0 +1,1165 @@
...
+
+export type Binding = [string, any] | ['Bind', number | string | any[]];
+export interface SqlResult {
+	query: string;
+	bindings: Binding[];
+}
+
+type MatchFn = (args: AbstractSqlQuery, indent: string) => string;
+
+const comparisons = {
```
zvin: this is already declared in AbstractSQLOptimiser.ts

AbstractSQLRules2SQL.ts
```diff
@@ -0,0 +1,1165 @@
...
+			}
+		}
+		return '$' + fieldOrderings.push(bind);
+	} else {
+		fieldOrderings.push(bind);
+		return '?';
+	}
+};
+
+let fieldOrderings: Binding[] = [];
```
thgreasi: Let's move this before `AddBind` so that it's not used before it's declaration

AbstractSQLRules2SQL.ts
```diff
@@ -0,0 +1,1165 @@
...
+let engine: Engines = Engines.postgres;
```
thgreasi: Let's move this one as well a bit higher, near the top of the file.

AbstractSQLRules2SQL.ts
```diff
@@ -0,0 +1,1165 @@
...
+					break;
+				case 'Limit':
+					if (limit !== '') {
+						throw new SyntaxError(
+							`'SelectQuery' can only accept one '${type}'`,
+						);
+					}
+					limit = indent + typeRules[type](rest, indent);
+					break;
+				case 'Offset':
```
thgreasi: Seems like Where, GroupBy, OrderBy, Limit & Offset could be using an abstract implementation

AbstractSQLRules2SQL.ts
```diff
@@ -0,0 +1,1165 @@
...
+				: '';
+
+		return (
+			'SELECT ' + select + from + where + groupBy + orderBy + limit + offset
+		);
+	},
+	Select: (args, indent) => {
+		checkArgs('Select', args, 1);
+		args = getAbstractSqlQuery(args, 0);
+		if (args.length === 0) {
```
thgreasi: Should we use `_.isEmpty()` to also protect from null/undefined?

AbstractSQLOptimiser.ts
```diff
@@ -0,0 +1,1217 @@
...
+	name: string,
+	matchers: MatchFn[],
+	rewriteFn: RewriteFn,
+): MatchFn => args => {
+	checkArgs(name, args, matchers.length);
+	return rewriteFn(
+		args.map((arg, index) => {
+			// if (!_.isArray(arg)) {
+			// 	throw new SyntaxError(`Expected AbstractSqlQuery array but got ${typeof arg}`)
+			// }
```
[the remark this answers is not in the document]
Page-: Nope, removed

AbstractSQLRules2SQL.ts
```diff
@@ -0,0 +1,1165 @@
...
+const CheckErr = (e: Error) => {
+	if (e instanceof SyntaxError || e instanceof TypeError) {
+		throw e;
+	}
+};
+
+const escapeField = (field: string | AbstractSqlQuery) =>
+	field === '*' ? '*' : `"${field}"`;
+
+const AnyValue: MatchFn = (args, indent) => {
```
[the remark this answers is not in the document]
Page-: Sort of, they should be equivalent (they operate on the same input) but one redirects the the optimisation functions and the other to the compilation functions, ie one returns an `AbstractSqlQuery` to go through more optimisation passes as needed or to then be passed to the compiler, whilst the other returns a SQL string

AbstractSQLRules2SQL.ts
Page-: `getAbstractSqlQuery` guarantees we're working with an array, so there's no need to guard against null/undefined

AbstractSQLCompiler.ts
```diff
@@ -1,3 +1,9 @@
+export enum Engines {
```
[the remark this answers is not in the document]
Page-: I checked and for this case and the object is still generated/exported (which is good) but it does change
```ts
let engine: Engines = Engines.postgres;
```
from
```js
var engine = AbstractSQLCompiler_1.Engines.postgres;
```
to
```js
var engine = "postgres";
```

AbstractSQLRules2SQL.ts
```diff
@@ -0,0 +1,1165 @@
...
+	GreaterThanOrEqual: ' >= ',
+	LessThan: ' < ',
+	LessThanOrEqual: ' <= ',
+	NotEquals: ' != ',
+	Like: ' LIKE ',
+};
+
+const NestedIndent = (indent: string): string => indent + '\t';
+
+const CheckErr = (e: Error) => {
```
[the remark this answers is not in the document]
Page-: For `comparisons` we only need its type in the optimiser, so I left it in here and just imported in the optimiser (the import isn't in the generated js), and for `CheckErr` with other changes it's now only used in one place so I just inlined it

Page-:
@zvin the typescript files aren't generated, but they are used for transforming the abstract sql and them compiling it to sql - so the sql the typescript version generates is the same as the ometajs version and so is compatible/only a minor change, but all of the typescript is handwritten and comments are definitely useful

Page-:
The main reason for the change was the (immensely) improved performance, but it also gives us better error messages (ie actual error messages with some information instead of just `match failed`) and a better ability for others to work with the code (since learning ometa takes a lot of time and doesn't have any real docs).
In terms of effort to work with the ometa vs typescript versions, I think for me the ometajs does win out slightly but only because I've had so much experience with it and that the abstract sql tree was largely designed to be easy for ometa to work with. And I think the accessibility for others is more important than that

AbstractSQLCompiler.ts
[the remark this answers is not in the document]
thgreasi: > the object is still generated/exported

Oh, that's probably b/c we use `"preserveConstEnums": true`:
https://github.com/balena-io-modules/abstract-sql-compiler/blob/59b932725f51ea68f3634beabd03f93d4190e279/tsconfig.json#L7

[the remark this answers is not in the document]
Page-: Ah, sweet, that makes life easier for non-ts interop :)

Page-:
I pushed some additional changes to avoid throwing errors as part of normal matching (throwing errors is expensive) and also deduplicated some code between compiler/optimiser (the `FooValue` checks)

Page-:
Not easily @dfunckt, it would in theory be possible to do something where you have the same traversal code for both and then different actions for the two, but passing around the indentation level makes that more awkward, as does the fact that the optimising step accepts more forms of AST than the compiler (largely due to backwards compatibility) which it then rewrites to forms the compiler accepts.  Definitely something that can be done but I think it would be a fairly significant piece of work atm

PR #46, opened 2018-11-20
Improve typings

Page-:
Change-type: minor


src/AbstractSQLCompiler.ts
```diff
@@ -15,6 +15,211 @@ import sbvrTypes = require('@resin/sbvr-types');
...
+	extends TwoArgNodeType<
+		'From',
+		| SelectQueryNode
+		| UnionQueryNode
+		| TableNode
+		| AliasNode<SelectQueryNode | UnionQueryNode | TableNode>
+	> {}
+export type TableNode = ['Table', string];
+export type WhereNode = ['Where', BooleanTypeNodes];
+// export type GroupByNode = ['GroupBy', Array<['ASC' | 'DESC', FieldNode | ReferencedFieldNode]>];
```
dfunckt: Is that commented out by accident?

src/AbstractSQLCompiler.ts
```diff
@@ -349,21 +553,25 @@ const compileSchema = (
 	ruleStatements = _.map(
 		abstractSqlModel.rules,
 		(rule): SqlRule => {
-			let ruleBody = _.find(rule, { 0: 'Body' } as Partial<AbstractSqlQuery>);
-			if (ruleBody == null || _.isString(ruleBody)) {
+			const ruleBodyNode = _.find(rule, { 0: 'Body' }) as [
+				'Body',
+				AbstractSqlQuery
+			];
```
dfunckt: Is it worth exporting this as `RuleBodyNode`?

src/AbstractSQLCompiler.ts
```diff
@@ -349,21 +553,25 @@ const compileSchema = (
...
 				throw new Error('Invalid rule');
 			}
-			let ruleSE = _.find(rule, { 0: 'StructuredEnglish' } as Partial<
-				AbstractSqlQuery
-			>);
-			if (ruleSE == null) {
+			const ruleSENode = _.find(rule, { 0: 'StructuredEnglish' }) as [
+				'StructuredEnglish',
+				string
+			];
```
dfunckt: This one as well

src/AbstractSQLOptimiser.ts
```diff
@@ -915,23 +893,42 @@ const typeRules: Dictionary<MatchFn> = {
 					throw new SyntaxError(`'InsertQuery' does not support '${type}'`);
 			}
 		}
+		if (tables.length === 0) {
+			throw new SyntaxError("'InsertQuery' must have a From");
```
dfunckt: This sounds a bit confusing -- I think it's better to change it like below to "InsertQuery requires a From component"

src/AbstractSQLOptimiser.ts
```diff
@@ -965,23 +962,35 @@ const typeRules: Dictionary<MatchFn> = {
...
+					where = [typeRules[type](rest)];
 					break;
 				default:
 					throw new SyntaxError(`'UpdateQuery' does not support '${type}'`);
 			}
 		}
-		if (fields.length !== values.length) {
-			throw new SyntaxError('Fields and Values must have the same length');
+		if (tables.length === 0) {
+			throw new SyntaxError("'UpdateQuery' must have a From");
```
dfunckt: Same here

src/AbstractSQLOptimiser.ts
```diff
@@ -995,21 +1004,24 @@ const typeRules: Dictionary<MatchFn> = {
...
 					}
-					where = typeRules[type](rest);
+					where = [typeRules[type](rest)];
 					break;
 				default:
 					throw new SyntaxError(`'DeleteQuery' does not support '${type}'`);
 			}
 		}
+		if (tables.length === 0) {
+			throw new SyntaxError("'DeleteQuery' must have a From");
```
dfunckt: Same here

src/AbstractSQLCompiler.ts
Page-: Kind of, it's meant to be a placeholder of "fixme"; I found out whilst adding types that I'd missed the groupby matcher as we don't actually use it anywhere, but I need to re-add it along with the typing

src/AbstractSQLCompiler.ts
Page-: I avoided exporting these two as they are a special case that only works for `abstractSqlModel.rules` and so we don't want them being typed as valid for the more common query case - they need to get covered in another pass where we eventually get rid of the generic `AbstractSqlQuery` type altogether

src/AbstractSQLOptimiser.ts
Page-: Makes sense

src/AbstractSQLCompiler.ts
Page-: Fixed

PR #57, opened 2019-05-23
Add support for computed fields

Page-:
Linked to https://github.com/balena-io-modules/odata-to-abstract-sql/pull/39 and https://github.com/balena-io/pinejs/pull/293

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -234,13 +234,14 @@ export interface AbstractSqlField {
 	fieldName: string;
 	dataType: string;
 	required?: boolean;
-	index: string;
+	index?: string;
 	references?: {
 		resourceName: string;
 		fieldName: string;
 	};
 	defaultValue?: string;
 	necessity: boolean;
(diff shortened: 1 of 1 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -234,13 +234,14 @@ export interface AbstractSqlField {
 	fieldName: string;
 	dataType: string;
 	required?: boolean;
-	index: string;
+	index?: string;
```
thgreasi: Any clue on why this became nullable?

Page-: It was always nullable and this was just typed incorrectly

Page-: As to why I changed it, when creating a computed field you need to be able to omit it and that creation is done in typescript so the types had to be correct (the place it's done for sbvr is https://github.com/balena-io-modules/lf-to-abstract-sql/blob/v3.0.1/lf-to-abstract-sql.ometajs which doesn't support typescript)

PR #58, opened 2019-05-30
Add tests for `getReferencedFields`

Page-:
Change-type: patch

### DIFF
--- a/test/abstract-sql/get-referenced-fields.ts
+++ b/test/abstract-sql/get-referenced-fields.ts
@@ -1,0 +1,305 @@
+import * as AbstractSqlCompiler from '../../src/AbstractSQLCompiler';
+import { expect } from 'chai';
+
+describe('getReferencedFields', () => {
+	it('should work with selected fields', () => {
+		expect(
+			AbstractSqlCompiler.postgres.getReferencedFields([
+				'SelectQuery',
+				[
+					'Select',
+					[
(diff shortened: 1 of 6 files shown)


test/abstract-sql/get-referenced-fields.ts
```diff
@@ -0,0 +1,204 @@
...
+						['ReferencedField', 'atable2', 'field1'],
+					],
+				],
+				['From', ['Alias', ['Table', 'table'], 'atable']],
+				['From', ['Alias', ['Table', 'table2'], 'atable2']],
+			]),
+		).to.deep.equal({
+			table: ['field1'],
+			atable: ['field1'],
+			table2: ['field1'],
```
dfunckt: I know I'm being somewhat pedantic here, but it would be better to assign different names to the fields since that would catch the case where the same list is accidentally referenced multiple times by the implementation.

test/abstract-sql/get-referenced-fields.ts
```diff
@@ -0,0 +1,204 @@
...
+						['ReferencedField', 'atable2', 'field2'],
+					],
+				],
+			]),
+		).to.deep.equal({
+			table: ['field1', 'field2'],
+			atable: ['field1'],
+			atable2: ['field2'],
+		});
+	});
```
dfunckt: It would be nice to have a test case that mixes fields from SELECT and WHERE.

PR #59, opened 2019-05-31
Use an abstract-sql-optimiser prepass to simplify getReferencedFields

Page-:
Change-type: patch

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -374,6 +374,8 @@ const dataTypeGen = (
 };
 
 const getReferencedFields: EngineInstance['getReferencedFields'] = ruleBody => {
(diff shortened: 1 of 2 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -397,26 +399,12 @@ const getReferencedFields: EngineInstance['getReferencedFields'] = ruleBody => {
...
-							if (_.isString(table)) {
-								// Deprecated: Remove this when we drop implicit tables
-								table = ['Table', table];
-							}
-							if (table[0] !== 'Table' || !_.isString(alias)) {
-								throw new Error('Cannot handle aliased select queries');
-							}
-							tableAliases[alias] = table[1];
+					if (nested[0] === 'Alias') {
+						let [, table, alias] = nested;
```
zvin: We can use a `const` here.

PR #61, opened 2019-06-14
 Add support for `IsDistinctFrom`/`IsNotDistinctFrom`

Page-:
This is split into two commits, one that has the prettier bump and `'postgres'`  -> `Engines.postgres` conversions, and the other that actually adds the new support, so you may want to review individually

### DIFF
--- a/src/AbstractSQLRules2SQL.ts
+++ b/src/AbstractSQLRules2SQL.ts
@@ -149,7 +149,9 @@ export const isBooleanValue = (
 		type === 'LessThan' ||
 		type === 'LessThanOrEqual' ||
 		type === 'NotEquals' ||
-		type === 'Like'
+		type === 'Like' ||
+		type === 'IsNotDistinctFrom' ||
+		type === 'IsDistinctFrom'
 	);
 };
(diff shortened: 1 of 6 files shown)


src/AbstractSQLRules2SQL.ts
```diff
@@ -622,6 +624,43 @@ const typeRules: Dictionary<MatchFn> = {
...
 	NotEquals: Comparison('NotEquals'),
 	Like: Comparison('Like'),
+	IsNotDistinctFrom: (args, indent) => {
+		checkArgs('IsNotDistinctFrom', args, 2);
+		const a = AnyValue(getAbstractSqlQuery(args, 0), indent);
+		const b = AnyValue(getAbstractSqlQuery(args, 1), indent);
+		if (engine === Engines.postgres) {
+			return a + ' IS NOT DISTINCT FROM ' + b;
+		} else if (engine === Engines.mysql) {
+			return a + ` <=> ` + b;
```
zvin: NIT: why are we using a backtick here?

Page-: A copy/paste thing I expect, will fix

PR #63, opened 2019-08-27
Add support for inner/left/right/full joins

Page-:
Change-type: minor


Page-:
You can, it's just rarely something you want to :P

PR #64, opened 2019-09-10
Optimise finding of existing binds when adding a bind variable

Page-:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,30 +19,30 @@
   "dependencies": {
     "@resin/sbvr-types": "^2.0.3",
     "@types/bluebird": "^3.5.27",
-    "@types/lodash": "^4.14.134",
-    "@types/node": "^8.10.49",
+    "@types/lodash": "^4.14.138",
+    "@types/node": "^8.10.53",
     "bluebird": "^3.5.5",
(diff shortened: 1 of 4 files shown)


roman-mazur:
Is it "inspired" by performance issues with service installs?

Page-:
Yup @roman-mazur, it was a case of a query that had a huge number of bind variables where suddenly the performance of the compare function really became noticeable, because it ones the compare function against each existing bind for each new bind added - not too bad with 10 or even 100 bind vars but definitely noticeable with 1000.  All this does is speed up that comparison function significantly in the common case so that it reduces the impact of that time complexity but it will still be a problem with enough bind vars. To handle massive numbers we'd need to change it to use a lookup table or similar instead to bring the time complexity down, but at the same time having that many bind vars by itself is a bit questionable

roman-mazur:
Great, thanks.
And do we collect any benchmark metrics using tests in `stress.coffee`?

Page-:
I did it manually and with `time npx mocha` for just stress.coffee it went from ~12.5s to ~4s and for my real world test case profiling it went from 15,241ms cpu time to 688ms

PR #65, opened 2019-09-12
Add support for resolving directly circular dependencies on postgres

Page-:
Change-type: minor

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -530,6 +530,7 @@ const compileSchema = (
 	}
 
 	const createSchemaStatements: string[] = [];
+	const alterSchemaStatements: string[] = [];
 	let dropSchemaStatements: string[] = [];
 
 	const fns: _.Dictionary<true> = {};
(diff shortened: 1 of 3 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -586,19 +587,46 @@ $$ LANGUAGE ${fnDefinition.language};`);
...
+					const fkDefinition = `FOREIGN KEY ("${fieldName}") REFERENCES "${referencedTable.name}" ("${references.fieldName}")`;
+
+					const schemaInfo = schemaDependencyMap[references.resourceName];
+					if (schemaInfo && schemaInfo.depends.includes(table.resourceName)) {
+						// It's a simple circular dependency so we switch it to an ALTER TABLE
+						// TODO: Remove the postgres specific "IF NOT EXISTS" check
+						alterSchemaStatements.push(`\
+DO $$
+BEGIN
+	IF NOT EXISTS (
```
roman-mazur: What are the cases this foreign key can already exist?

dfunckt:
Do you have a real need for this one @Page- ? It feels to me it's enabling a use case that is problematic in the first place and should be solved at a higher level.

Page-:
Yes @dfunckt: https://github.com/balena-io/open-balena-api/pull/163

src/AbstractSQLCompiler.ts
Page-: The create schema statements are rerun on pinejs startup to create any new tables that are missing (or in the case of a new env to create everything) so in those cases the foreign keys can already exist and is why all of the create statements include the `IF NOT EXISTS` equivalent

PR #79, opened 2020-02-21
Improve typings

Page-:
Change-type: patch

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -28,104 +28,31 @@ export type DurationNode = [
 	},
 ];
 
-// The extends array hacks in the node types are because otherwise we get issues with circular refs
-export interface OneArgNodeType<T, X> extends Array<T | X> {
-	0: T;
-	1: X;
-	length: 2;
-}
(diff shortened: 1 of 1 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -148,21 +74,23 @@ export type NumberTypeNodes = NumberNode | CountNode | UnknownTypeNodes;
 export type FieldNode = ['Field', string];
 export type ReferencedFieldNode = ['ReferencedField', string, string];
 export type BindNode = ['Bind', string, string?];
-export interface CastNode
-	extends TwoVarArgNodeType<'Cast', AbstractSqlType, string> {}
+export type CastNode = ['Cast', AbstractSqlType?, ...string[]];
```
dfunckt: Is this surely an `AbstractSqlType?`? It was previously just `AbstractSqlType`.

Page-: It was typed as `AbstractSqlType?` previously, but in reality it should be `['Cast', AbstractSqlType, string]` so I fixed both parts

Page-:
Merged into https://github.com/balena-io-modules/abstract-sql-compiler/pull/88

PR #88, opened 2020-05-19
7.x major version

Page-:
* Drop support for node 6/8
* Update tsconfig to target es2018
* Convert all returned promises to native promises instead of bluebird
* Update to @balena/sbvr-types 3.0.0
* Improve typings - this is https://github.com/balena-io-modules/abstract-sql-compiler/pull/79 but brought into this as it is a major change in terms of typings
* Rename to @balena/abstract-sql-compiler

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -11,7 +11,7 @@ import {
 	SqlResult,
 } from './AbstractSQLRules2SQL';
 export { Binding, SqlResult } from './AbstractSQLRules2SQL';
-import sbvrTypes = require('@resin/sbvr-types');
+import sbvrTypes = require('@balena/sbvr-types');
 import * as _ from 'lodash';
 
 export type NullNode = ['Null'];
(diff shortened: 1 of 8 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -374,14 +315,17 @@ export interface EngineInstance {
 
 const validateTypes = _.mapValues(sbvrTypes, ({ validate }) => validate);
 
-const dataTypeValidate: EngineInstance['dataTypeValidate'] = (value, field) => {
+const dataTypeValidate: EngineInstance['dataTypeValidate'] = async (
```
thgreasi, line 318: I think this doesn't need to be an `async `function, since I remember that we changes the validation functions in the sbvr types to synchronous code.
:thinking: that it's just the typings that are wrong in there.

thgreasi: No we didn't, since `runCheck` is still marked as async...

PR #91, opened 2020-10-02
Remove accidental `necessity` from the `AbstractSqlField` types

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

dfunckt:
This should be a major version bump. In fact the cloud API is using `necessity` and the next build is going to fail.

Page-:
@dfunckt using the property is a bug and won't actually do anything you might expect it to, as such I count this as a bugfix - and like any other bugfix if you rely on the bugged behavior then it is a breaking change, but relying on that behavior is a sign of an issue that you should be prompted to fix imo

PR #92, opened 2020-11-25
Update sbvr-types and make use of `castType`

Page-:
Update sbvr-types from 3.1.0 to 3.1.2

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #95, opened 2020-11-30
Add an `optimizeSchema` function

Page-:
Currently this can optimize certain rules into check constraints


Page-:
@balena-ci I self-certify!

PR #100, opened 2020-12-18
Add support for table definitions

Page-:
Static definitions will be converted to views and dynamic definitions
will be ignored

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #101, opened 2020-12-18
Fix `containsNode` when that node type is at the top level

Page-:
Change-type: patch


Page-:
@balena-ci rebase

Page-:
@balena-ci I self-certify!

Page-:
@balena-ci retest

PR #102, opened 2020-12-21
Improve Relationship typings

thgreasi:
Change-type: patch
See: https://github.com/balena-io/open-balena-api/pull/510#discussion_r546872607
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -294,17 +294,19 @@ export interface SqlRule {
  *   - RelationshipMapping[1][1] is the name of the field on the other resource
  */
 export type RelationshipMapping = [string, [string, string]?];
-export interface Relationship {
+export interface RelationshipLeafNode {
 	$: RelationshipMapping;
-	// TODO: This should action be just Relationship, but we can't declare that in typescript currently
-	[resourceName: string]: Relationship | RelationshipMapping;
 }
+export interface RelationshipInternalNode {
+	[resourceName: string]: Relationship;
+}
+export type Relationship = RelationshipLeafNode | RelationshipInternalNode;
 export interface AbstractSqlModel {
 	synonyms: {
 		[synonym: string]: string;
 	};
 	relationships: {
-		[resourceName: string]: Relationship;
+		[resourceName: string]: RelationshipInternalNode;
 	};
 	tables: {
 		[resourceName: string]: AbstractSqlTable;


src/AbstractSQLCompiler.ts
```diff
@@ -294,17 +294,19 @@ export interface SqlRule {
...
-export interface Relationship {
+export interface RelationshipLeafNode {
 	$: RelationshipMapping;
-	// TODO: This should action be just Relationship, but we can't declare that in typescript currently
-	[resourceName: string]: Relationship | RelationshipMapping;
 }
+export interface RelationshipInternalNode {
+	[resourceName: string]: Relationship;
+}
+export type Relationship = RelationshipLeafNode | RelationshipInternalNode;
```
Page-, line 303: Do you know how this works with the fact that internal and leaf nodes can be combined? eg from the user model in @balena/pinejs you get:
```js
"user-has-permission": {
	"user": {
		"$": [
			"user",
			[
				"user",
				"id"
			]
		],
		"has": {
			"permission": {
				"has": {
					"expiry date": {
						"$": [
							"expiry date"
						]
					}
				}
			}
		}
	},
```

thgreasi: You can assign that normally to a Relationship and I actually got the relationships object from the abstract model that the API produces and TS didn't complain assigning it here.
When I tried the same on master I was getting errors really soon, like:
```
Type '{ Integer: { $: [string]; }; }' is not assignable to type 'RelationshipMapping | Relationship'.
  Property '$' is missing in type '{ Integer: { $: [string]; }; }' but required in type 'Relationship'.

(property) has: {
    Integer: {
        $: [string];
    };
}
```

Here are the API changes that this requires:
See: https://github.com/balena-io/balena-api/pull/2920

We will now have to add an extra `Arrayl.isArray(x.$)`, which was previously 
```

const relationships: AbstractSqlModel['relationships'] = {...};
const userRelationship = relationships['user-has-permission'].user;

// @ts-expect-error
console.log(userRelationship.$[1]);
// @ts-expect-error
console.log(userRelationship.has);

// working
console.log(userRelationship.$); // before this PR this was complaining as well
console.log(Array.isArray(userRelationship.$) && userRelationship.$[1]);
console.log('has' in userRelationship && userRelationship.has); // you had to that before this PR as well
```

Related upsream PRs:
See: https://github.com/balena-io/pinejs/pull/429
See: https://github.com/balena-io-modules/odata-to-abstract-sql/pull/70

Page- (approved):

PR #103, opened 2021-01-01
Optimize out booleans from AND/OR where possible

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #104, opened 2021-01-01
Add the ability to optimize generated sql using model info

Page-:
We automatically use the model info to optimize rules/check constraints
when compiling the schema. Currently it supports optimizing exists
checks for referenced field nodes.

Change-type: minor


Page-:
@balena-ci retest

test/sbvr/pilots.js
```diff
@@ -215,6 +223,12 @@ SELECT NOT EXISTS (
 	test.rule(
 		'It is necessary that each plane that at least 3 pilots can fly, has a name',
 		`\
+SELECT 1 AS "result";`,
```
Page-, line 226: This is what it looks like when you write a rule that cannot be broken :P

test/sbvr/pilots.js
```diff
@@ -26,6 +26,8 @@ describe('pilots', function () {
 		`\
 Term:      name
 	Concept Type: Short Text (Type)
+Term:      nickname
+	Concept Type: Short Text (Type)
```
Page-, line 30: I needed a nullable term to actually be able to test we generated sensible rules as opposed to having them all optimized away to nothing because they couldn't be broken :|

Page-:
@balena-ci retest

PR #105, opened 2021-01-01
Optimize away AND/ORs with false/true respectively

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #109, opened 2021-01-15
Fix schema optimizations when table and resource names are different

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #110, opened 2021-01-25
Optimize the abstract sql of rules when running `optimizeSchema`

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #112, opened 2021-01-29
Add a `getRuleReferencedFields` which is tailored for rules

Page-:
This works on the assumption that it is a rule query that should always
return true and allows ignoring more modification cases where they
cannot change the result from true to false

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #115, opened 2021-03-01
Generalize/share the referenced fields code and cover more cases

Page-:
This now allows detecting deletes that are

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #119, opened 2021-05-03
Remove now unnecessary hack for casting SERIAL types

Page-:
This became unnecessary when @balena/sbvr-types added the concept of
cast types since that now specifies how to cast a SERIAL type

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #128, opened 2021-07-26
Supprt date_trunc postgres date function

fisehara:
Supprt date_trunc postgres date function
for DateValue

Change-type: patch
Signed-off-by: fisehara harald@balena.io

### DIFF
--- a/src/AbstractSQLOptimiser.ts
+++ b/src/AbstractSQLOptimiser.ts
@@ -650,6 +650,7 @@ const typeRules: Dictionary<MatchFn> = {
 	Floor: matchArgs('Floor', NumericValue),
 	Ceiling: matchArgs('Ceiling', NumericValue),
 	ToDate: matchArgs('ToDate', DateValue),
+	DateTrunc: matchArgs('DateTrunc', TextValue, DateValue),
 	ToTime: matchArgs('ToTime', DateValue),
 	Coalesce: (args) => {
 		checkMinArgs('Coalesce', args, 2);
--- a/src/AbstractSQLRules2SQL.ts
+++ b/src/AbstractSQLRules2SQL.ts
@@ -178,7 +178,8 @@ export const isDateValue = (
 		type === 'ToDate' ||
 		type === 'ToTime' ||
 		type === 'CurrentTimestamp' ||
-		type === 'CurrentDate'
+		type === 'CurrentDate' ||
+		type === 'DateTrunc'
 	);
 };
 const DateValue = MatchValue(isDateValue);
@@ -938,6 +939,15 @@ const typeRules: Dictionary<MatchFn> = {
 		checkArgs('ToDate', args, 1);
 		const date = DateValue(getAbstractSqlQuery(args, 0), indent);
 		return `DATE(${date})`;
+	},
+	DateTrunc: (args, indent) => {
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError('DateTrunc is not supported on: ' + engine);
+		}
+		checkArgs('DateTrunc', args, 2);
+		const precision = TextValue(getAbstractSqlQuery(args, 0), indent);
+		const date = DateValue(getAbstractSqlQuery(args, 1), indent);
+		return `DATE_TRUNC(${precision}, ${date})`;
 	},
 	ToTime: (args, indent) => {
 		checkArgs('ToTime', args, 1);
--- a/test/abstract-sql/functions_wrapper.ts
+++ b/test/abstract-sql/functions_wrapper.ts
@@ -1,0 +1,51 @@
+import { stripIndent } from 'common-tags';
+import { AbstractSqlQuery } from '../../src/AbstractSQLCompiler';
+
+type TestCb = (
+	result: { query: string },
+	sqlEquals: (a: string, b: string) => void,
+) => void;
+// tslint:disable-next-line no-var-requires
+const test = require('./test') as (
+	query: AbstractSqlQuery,
+	binds: any[][] | TestCb,
+	cb?: TestCb,
+) => void;
+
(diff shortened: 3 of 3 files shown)


src/AbstractSQLOptimiser.ts
```diff
@@ -650,6 +650,7 @@ const typeRules: Dictionary<MatchFn> = {
 	Floor: matchArgs('Floor', NumericValue),
 	Ceiling: matchArgs('Ceiling', NumericValue),
 	ToDate: matchArgs('ToDate', DateValue),
+	DateTrunc: matchArgs('DateTrunc', Field, _.identity),
```
Page-, line 653: I think this can operate on any date value and not just fields right?
```suggestion
	DateTrunc: matchArgs('DateTrunc', DateValue, _.identity),
```
And the precision value needs to be text right, so it can have improved validation via
```suggestion
	DateTrunc: matchArgs('DateTrunc', DateValue, TextValue),
```
Also I think it'd be best to keep the order the same as the postgres function since it's basically a 1-1 mapping and having it switched could make it more confusing to use (although I agree in a vacuum I'd also go `date, precision` rather than `precision, date`)
```suggestion
	DateTrunc: matchArgs('DateTrunc', TextValue, DateValue),
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -939,6 +940,12 @@ const typeRules: Dictionary<MatchFn> = {
 		const date = DateValue(getAbstractSqlQuery(args, 0), indent);
 		return `DATE(${date})`;
 	},
+	DateTrunc: (args, indent) => {
+		checkArgs('DateTrunc', args, 2);
+		const field = Field(getAbstractSqlQuery(args, 0), indent);
+		const str = args[1];
+		return `date_trunc('${str}',${field})`;
+	},
```
Page-, line 948: Can you add an
```ts
if (engine === Engines.postgres) {
```
check please and throw an error otherwise, since afaict only postgres supports the DATE_TRUNC function and not websql/mysql

src/AbstractSQLRules2SQL.ts
```diff
@@ -939,6 +940,12 @@ const typeRules: Dictionary<MatchFn> = {
 		const date = DateValue(getAbstractSqlQuery(args, 0), indent);
 		return `DATE(${date})`;
 	},
+	DateTrunc: (args, indent) => {
+		checkArgs('DateTrunc', args, 2);
+		const field = Field(getAbstractSqlQuery(args, 0), indent);
+		const str = args[1];
+		return `date_trunc('${str}',${field})`;
```
Page-, line 947: To match the format of the rest of the generated sql:
```suggestion
		return `DATE_TRUNC('${str}', ${field})`;
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -939,6 +940,12 @@ const typeRules: Dictionary<MatchFn> = {
 		const date = DateValue(getAbstractSqlQuery(args, 0), indent);
 		return `DATE(${date})`;
 	},
+	DateTrunc: (args, indent) => {
+		checkArgs('DateTrunc', args, 2);
+		const field = Field(getAbstractSqlQuery(args, 0), indent);
+		const str = args[1];
```
Page-, line 946: This 100% needs some form of validation rather than just dumping it into the sql as-is as that opens up all sorts of potential issues, especially if it comes directly from a user, using
```suggestion
		const str = TextValue(args[1], indent);
```
should be able to handle that

src/AbstractSQLRules2SQL.ts
```diff
@@ -939,6 +940,17 @@ const typeRules: Dictionary<MatchFn> = {
...
+	DateTrunc: (args, indent) => {
+		if (engine !== Engines.postgres) {
+			throw new Error(
+				'DATE_TRUNC function is only supported on postgres currently',
+			);
+		}
+		checkArgs('DateTrunc', args, 2);
+		const str = TextValue(getAbstractSqlQuery(args, 0), indent);
+		const field = DateValue(getAbstractSqlQuery(args, 1), indent);
+		return `DATE_TRUNC(${str}, ${field})`;
```
Page-, line 952: I'd recommend using names that match more closely to what they are so it's easier to know without looking at docs:
```suggestion
		const precision = TextValue(getAbstractSqlQuery(args, 0), indent);
		const date = DateValue(getAbstractSqlQuery(args, 1), indent);
		return `DATE_TRUNC(${precision}, ${date})`;
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -939,6 +940,17 @@ const typeRules: Dictionary<MatchFn> = {
 		const date = DateValue(getAbstractSqlQuery(args, 0), indent);
 		return `DATE(${date})`;
 	},
+	DateTrunc: (args, indent) => {
+		if (engine !== Engines.postgres) {
+			throw new Error(
+				'DATE_TRUNC function is only supported on postgres currently',
+			);
```
Page-, line 947: I'd make this a `SyntaxError` and specify the engine it's not supported on in order to match the others
```suggestion
			throw new SyntaxError('DateTrunc is not supported on: ' + engine);
```

test/abstract-sql/functions_wrapper.ts
```diff
@@ -0,0 +1,52 @@
...
+				],
+			],
+		],
+		[
+			['Text', 'milliseconds'],
+			['Bind', 0],
+		],
+		(result, sqlEquals) => {
+			it('Generate a postgresql DATE_TRUNC query for referenced field with milliseconds resoluton and binding', () => {
+				console.log(JSON.stringify(result, null, 2));
```
Page-, line 40: Leftover debug info

Page- (approved):

PR #132, opened 2021-09-13
Relax the restrictions on `ORDER BY` as it doesn't just allow fields

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #139, opened 2021-12-14
Avoid the operation type optimization for HAVING clauses

Page-:
Since it's possible for them to behave in a way that violates the
safety checks we'll disable them until we can add specific checks
for the HAVING clause

Change-type: patch


Page-:
@balena-ci I self-certify!

Page-:
@balena-ci rebase

PR #140, opened 2022-03-22
Add support for `ConcatenateWithSeparator` nodes

thgreasi:
Change-type: minor


src/AbstractSQLRules2SQL.ts
```diff
@@ -864,6 +865,23 @@ const typeRules: Dictionary<MatchFn> = {
...
+				);
+			}
+			return TextValue(arg, indent);
+		});
+		if (engine === Engines.websql) {
+			throw new SyntaxError(
+				'ConcatenateWithSeparator not supported on: ' + engine,
+			);
+		}
+		return `CONCAT_WS(${separator}, ${textParts.join(', ')})`;
```
Page-, line 883: It's not strictly necessary to separate out the separator, eg
```suggestion
		const textParts = args.map((arg) => {
			if (!isAbstractSqlQuery(arg)) {
				throw new SyntaxError(
					`Expected AbstractSqlQuery array but got ${typeof arg}`,
				);
			}
			return TextValue(arg, indent);
		});
		if (engine === Engines.websql) {
			throw new SyntaxError(
				'ConcatenateWithSeparator not supported on: ' + engine,
			);
		}
		return `CONCAT_WS(${textParts.join(', ')})`;
```

thgreasi: Fair point

PR #143, opened 2022-06-06
Add `modifyFields`/`modifyName` to table definition

Page-:
This is intended to be used to target tables differently for reads vs
writes, eg where you might read from a view/definition but want to
write to a physical table instead

Change-type: minor

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -349,6 +349,8 @@ export interface AbstractSqlTable {
 	triggers?: Trigger[];
 	checks?: Check[];
 	definition?: Definition;
+	modifyFields?: AbstractSqlTable['fields'];
+	modifyName?: AbstractSqlTable['name'];
 }
 export interface SqlRule {
 	sql: string;


thgreasi (commented):
Not sure whether `writeFields` would sounds better (since that's also how described it in the odata-to-abstract-sql PR) but LGTM :+1:

fisehara (commented):
I'm fine with modifyFields.
I'd like to understand why we would not name it translatedFields and translatedName when we use it for this solely right in the moment? It's something to 'know' that modifyFields is needed for the pinejs translations concept. 

Otherwise, LGTM.

thgreasi:
@fisehara as per the description this set of field definitions when defined are used or all write operations on the resources.

Page-:
Yeah, this isn't specifically for translations as it could be used for other purposes as well (eg writable view type thing) hence why I named it generically even if it's main use case at first will be translations. For reference there's definitely been other translations related work that was merged in the past that subsequent grew into many other use cases

Page-:
@balena-ci I self-certify!

PR #145, opened 2022-11-09
Add ToJSON/ExtractJSONPathAsText functions

joshbwlng:
Change-type: minor

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -103,15 +103,18 @@ export type CoalesceNode = [
 	UnknownTypeNodes,
 	...UnknownTypeNodes[],
 ];
+export type ToJSONNode = ['ToJSON', AnyTypeNodes];
 export type UnknownTypeNodes =
 	| FieldNode
 	| ReferencedFieldNode
 	| BindNode
 	| CastNode
 	| CoalesceNode
+	| ToJSONNode
 	| UnknownNode;
 
 export type TextNode = ['Text', string];
+export type JSONNode = ['JSON', string];
 export type ConcatenateNode = ['Concatenate', ...TextTypeNodes[]];
 export type ConcatenateWithSeparatorNode = [
 	'ConcatenateWithSeparator',
@@ -125,11 +128,19 @@ export type ReplaceNode = [
 	TextTypeNodes,
 	TextTypeNodes,
 ];
+export type ExtractJSONPathAsTextNode = [
+	'ExtractJSONPathAsText',
+	JSONNode,
+	TextArrayTypeNodes,
+];
+export type TextArrayTypeNodes = TextArrayNode | UnknownNode;
+export type TextArrayNode = ['TextArray', ...TextNode[]];
 export type TextTypeNodes =
 	| ConcatenateNode
 	| ConcatenateWithSeparatorNode
 	| LikeNode
 	| ReplaceNode
+	| ExtractJSONPathAsTextNode
 	| UnknownTypeNodes;
 
 export type SelectQueryNode = [
@@ -213,12 +224,12 @@ export type ValuesNodeTypes =
 	| NullNode
 	| BindNode
 	| TextNode
+	| JSONNode
 	| NumberNode;
 
 export type AliasNode<T> = ['Alias', T, string];
 
-export type AbstractSqlType =
-	| string
+export type AnyTypeNodes =
 	| NullNode
 	| DateNode
 	| BooleanTypeNodes
@@ -241,6 +252,9 @@ export type AbstractSqlType =
 	| GroupByNode
 	| HavingNode
 	| UnknownNode;
+
+export type AbstractSqlType = string | AnyTypeNodes;
+
 export type UnknownNode = AbstractSqlQuery;
 export interface AbstractSqlQuery extends Array<AbstractSqlType> {
 	0: string;
--- a/src/AbstractSQLOptimiser.ts
+++ b/src/AbstractSQLOptimiser.ts
@@ -137,6 +137,8 @@ const UnknownValue: MetaMatchFn = (args) => {
 		case 'Bind':
 		case 'Cast':
 		case 'Coalesce':
+		case 'ExtractJSONPathAsText':
+		case 'ToJSON':
 			return typeRules[type](rest);
 		case 'SelectQuery':
 		case 'UnionQuery':
@@ -196,6 +198,9 @@ const { isJSONValue } = AbstractSQLRules2SQL;
 
 const { isDurationValue } = AbstractSQLRules2SQL;
 const DurationValue = MatchValue(isDurationValue);
+
+const { isArrayValue } = AbstractSQLRules2SQL;
+const ArrayValue = MatchValue(isArrayValue);
 
 const { isFieldValue } = AbstractSQLRules2SQL;
 const Field: MetaMatchFn = (args) => {
@@ -672,6 +677,17 @@ const typeRules: Dictionary<MatchFn> = {
 	ToDate: matchArgs('ToDate', DateValue),
 	DateTrunc: matchArgs('DateTrunc', TextValue, DateValue),
 	ToTime: matchArgs('ToTime', DateValue),
+	ExtractJSONPathAsText: (args) => {
+		checkMinArgs('ExtractJSONPathAsText', args, 1);
+		const json = TextValue(getAbstractSqlQuery(args, 0));
+		const path = ArrayValue(getAbstractSqlQuery(args, 1));
+		return ['ExtractJSONPathAsText', json, path];
+	},
+	TextArray: (args) => {
+		// Allow for populated and empty arrays
+		return ['TextArray', ...args.map(TextValue)];
+	},
+	ToJSON: matchArgs('ToJSON', AnyValue),
 	Coalesce: (args) => {
 		checkMinArgs('Coalesce', args, 2);
 		return ['Coalesce', ...args.map(AnyValue)];
--- a/src/AbstractSQLRules2SQL.ts
+++ b/src/AbstractSQLRules2SQL.ts
@@ -77,6 +77,8 @@ const UnknownValue: MetaMatchFn = (args, indent) => {
 		case 'Bind':
 		case 'Cast':
 		case 'Coalesce':
+		case 'ToJSON':
+		case 'TextArray':
 			return typeRules[type](rest, indent);
 		case 'SelectQuery':
 		case 'UnionQuery':
@@ -109,6 +111,7 @@ export const isTextValue = (
 		type === 'Upper' ||
 		type === 'Trim' ||
 		type === 'Replace' ||
+		type === 'ExtractJSONPathAsText' ||
 		type === 'Substring' ||
 		type === 'Right'
 	);
@@ -184,12 +187,18 @@ export const isDateValue = (
 	);
 };
 const DateValue = MatchValue(isDateValue);
+export const isArrayValue = (
+	type: string | AbstractSqlQuery,
+): type is string => {
+	return type === 'TextArray';
+};
 
 export const isJSONValue = (
 	type: string | AbstractSqlQuery,
 ): type is string => {
-	return type === 'AggregateJSON';
+	return type === 'AggregateJSON' || type === 'ToJSON';
 };
+const JSONValue = MatchValue(isJSONValue);
 
 export const isDurationValue = (
 	type: string | AbstractSqlQuery,
@@ -782,6 +791,19 @@ const typeRules: Dictionary<MatchFn> = {
 		checkArgs('EmbeddedText', args, 1);
 		return `'${args[0]}'`;
 	},
+	TextArray: (args, indent) => {
+		const values = args.map((arg) => {
+			if (!isAbstractSqlQuery(arg)) {
+				throw new SyntaxError(
+					`Expected AbstractSqlQuery array but got ${typeof arg}`,
+				);
+			}
+			return TextValue(arg, indent);
+		});
+		return values.length
+			? `ARRAY[${values.join(', ')}]`
+			: 'CAST(ARRAY[] as TEXT[])';
+	},
 	Null: (args) => {
 		checkArgs('Null', args, 0);
 		return 'NULL';
@@ -889,6 +911,17 @@ const typeRules: Dictionary<MatchFn> = {
 		const replacement = TextValue(getAbstractSqlQuery(args, 2), indent);
 		return `REPLACE(${str}, ${find}, ${replacement})`;
 	},
+	ExtractJSONPathAsText: (args, indent) => {
+		checkMinArgs('ExtractJSONPathAsText', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError(
+				'ExtractJSONPathAsText not supported on: ' + engine,
+			);
+		}
+		const json = JSONValue(getAbstractSqlQuery(args, 0), indent);
+		const path = TextValue(getAbstractSqlQuery(args, 1), indent);
+		return `${json} #>> ${path}`;
+	},
 	CharacterLength: (args, indent) => {
 		checkArgs('CharacterLength', args, 1);
 		const text = TextValue(getAbstractSqlQuery(args, 0), indent);
@@ -995,6 +1028,14 @@ const typeRules: Dictionary<MatchFn> = {
 		} else {
 			return `TIME(${date})`;
 		}
+	},
+	ToJSON: (args, indent) => {
+		checkMinArgs('ToJSON', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError('ToJSON not supported on: ' + engine);
+		}
+		const value = AnyValue(getAbstractSqlQuery(args, 0), indent);
+		return `TO_JSON(${value})`;
 	},
 	Coalesce: (args, indent) => {
 		checkMinArgs('Coalesce', args, 2);
--- a/test/abstract-sql/json.ts
+++ b/test/abstract-sql/json.ts
@@ -1,0 +1,72 @@
+import { AbstractSqlQuery } from '../../src/AbstractSQLCompiler';
+
+type TestCb = (
+	result: { query: string },
+	sqlEquals: (a: string, b: string) => void,
+) => void;
+// tslint:disable-next-line no-var-requires
+const test = require('./test') as (
+	query: AbstractSqlQuery,
+	binds: any[][] | TestCb,
+	cb?: TestCb,
+) => void;
+
+describe('ExtractJSONPathAsText', () => {
+	test(
+		[
+			'SelectQuery',
+			[
+				'Select',
+				[
+					[
+						'ExtractJSONPathAsText',
+						['ReferencedField', 'foo', 'bar'],
+						['TextArray'],
+					],
+				],
+			],
+		],
+		(result, sqlEquals) => {
+			it('should produce a valid select with given empty text array', () => {
+				sqlEquals(
+					result.query,
+					`SELECT "foo"."bar" #>> CAST(ARRAY[] as TEXT[])`,
+				);
+			});
+		},
+	);
+	test(
+		[
+			'SelectQuery',
+			[
+				'Select',
+				[
+					[
+						'ExtractJSONPathAsText',
+						['ReferencedField', 'foo', 'bar'],
+						['TextArray', ['EmbeddedText', 'buz'], ['EmbeddedText', 'baz']],
+					],
+				],
+			],
+		],
+		(result, sqlEquals) => {
+			it('should produce a valid select with given populated text array', () => {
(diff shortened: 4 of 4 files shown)


joshbwlng:
@balena-ci rebase

src/AbstractSQLRules2SQL.ts
```diff
@@ -889,6 +891,14 @@ const typeRules: Dictionary<MatchFn> = {
...
 		return `REPLACE(${str}, ${find}, ${replacement})`;
 	},
+	ParseJSON: (args, indent) => {
+		checkArgs('ParseJSON', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError('ParseJSON not supported on: ' + engine);
+		}
+		const column = TextValue(getAbstractSqlQuery(args, 0), indent);
+		return `${column}::jsonb ->> 0`;
+	},
```
Page-, line 901: Same thing here, this isn't strictly parsing JSON but rather extracting an elememt as text, so eg `->` is `ExtractJSON` and `->>` is `ExtractJSONText` (the extra `>` in all these functions means to return text instead of JSON). But I found an even better way which is to use `#>>'{}'`. Overall I think it'd be best to do an
```
['ExtractJSONPath', JSONNode, path: TextNode]
```
although I'm not sure on
```
['ExtractJSONPath', JSONNode, path: TextNode, returnText?: boolean]
```
or
```
['ExtractJSONPathAsText', JSONNode, path: TextNode]
```
I'm leaning towards the latter with `ExtractJSONPathAsText` because it makes things like typechecking easier because
```
`ExtractJSONPathAsTextNode` would be a `TextNode` and `ExtractJSONPath` would be a `JSONNode` but the alternative is that `ExtractJSONPath extends [any, any, any, true]` is a `TextNode` but `ExtractJSONPath extends [any, any, any, false]` is a `JSONNode` and I expect it'll make things a pain

File: test/abstract-sql/stringify-json.ts [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+
+describe('StringifyJSON', () => {
+	test(
+		[
+			'SelectQuery',
+			['Select', [['StringifyJSON', ['ReferencedField', 'foo', 'bar']]]],
+		],
+		(result, sqlEquals) => {
+			it('should produce a valid json to text select statement', () => {
+				console.log('result:', result);
```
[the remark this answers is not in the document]
joshbwlng: Yep, removed

test/abstract-sql/parse-json.ts
```diff
@@ -0,0 +1,27 @@
...
+
+describe('ParseJSON', () => {
+	test(
+		[
+			'SelectQuery',
+			['Select', [['ParseJSON', ['ReferencedField', 'foo', 'bar']]]],
+		],
+		(result, sqlEquals) => {
+			it('should produce a valid jsonb first element select statement', () => {
+				console.log('result:', result);
```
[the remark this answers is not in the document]
joshbwlng: Yep, removed

src/AbstractSQLRules2SQL.ts
```diff
@@ -996,6 +1006,14 @@ const typeRules: Dictionary<MatchFn> = {
 			return `TIME(${date})`;
 		}
 	},
+	StringifyJSON: (args, indent) => {
```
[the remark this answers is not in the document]
joshbwlng: Fair point. I was working from the concept of `JSON.parse()` and `JSON.stringify()` which is why they are named the way they are, but I agree there's no reason to stick with those names if they technically don't mean the same things.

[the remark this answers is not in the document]
joshbwlng: Especially since what this does is just call `to_json()` if Postgres

src/AbstractSQLRules2SQL.ts
joshbwlng: I agree with being specific with the types, `ExtractJSONPathAsText` in this case. There will probably be more cases in the future where we might want `ExtractJSONPathAsJSON`, `ExtractJSONPathAsNumber`, etc. I'm thinking it would be pretty difficult to try and cover all cases with a single `ExtractJSONPath` and makes sense to manage return types at the function layer and not a number of extra parameters.

joshbwlng: @Page- Taking another look at the [Postgres docs](https://www.postgresql.org/docs/14/functions-json.html), I think we should just go ahead and call `json_extract_path_text()`. According to the docs, its equivalent to the `#>>` operator. Also the name makes it more obvious/explicit as to what we're doing, almost exactly the same name as our wrapper here.

joshbwlng: Sounds like we might should just call `json_extract_path_text()` if we're allowing callers to specify a path.

joshbwlng: Although `json_extract_pah_text()` seems to require a path passed in while that won't always be the case for us. In your example above, I'm think it should be `['ExtractJSONPathAsText', JSONNode, path?: TextNode]`, with the `path` being optional instead of required. When not provided, the entire JSON is returned as text.

src/AbstractSQLCompiler.ts
```diff
@@ -103,15 +103,18 @@ export type CoalesceNode = [
...
 	| FieldNode
 	| ReferencedFieldNode
 	| BindNode
 	| CastNode
 	| CoalesceNode
+	| ToJSONNode
 	| UnknownNode;
 
 export type TextNode = ['Text', string];
+export type JSONNode = ['JSON', string];
```
joshbwlng, line 117: Not sure how to define `JSONNode`, from what I understand Postgres treats `JSON` columns roughly the same as `TEXT`, just with basic JSON validation.

src/AbstractSQLCompiler.ts
```diff
@@ -213,9 +222,11 @@ export type ValuesNodeTypes =
 	| NullNode
 	| BindNode
 	| TextNode
+	| JSONNode
 	| NumberNode;
 
 export type AliasNode<T> = ['Alias', T, string];
+export type JSONPathNode = ['JSONPath', string];
```
joshbwlng, line 229: Could possibly handle `string[]` instead of `string` and `join()` during SQL generation. Feels like it could be nicer, but probably wouldn't be portable if/when `ExtractJSONPathAsText` is extended to work with MySQL/SQLite.

src/AbstractSQLRules2SQL.ts
```diff
@@ -889,6 +892,17 @@ const typeRules: Dictionary<MatchFn> = {
...
 	},
+	ExtractJSONPathAsText: (args, indent) => {
+		checkMinArgs('ExtractJSONPathAsText', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError(
+				'ExtractJSONPathAsText not supported on: ' + engine,
+			);
+		}
+		const column = TextValue(getAbstractSqlQuery(args, 0), indent);
+		const path = args[1] ? getAbstractSqlQuery(args, 1)[1] : '';
```
joshbwlng, line 903: `getAbstractSqlQuery(args, 1)[1]` feels hacky...

Page-: Yeah, really it should be something like
```suggestion
		const column = TextArrayValue(getAbstractSqlQuery(args, 1), indent);
```
since `{foo}` is just an array literal and can be supplied in different forms, eg
```sql
SELECT '{"foo":{"bar":"baz"}}'::json #>> ARRAY_AGG(t.path), ARRAY_AGG(t.path)
FROM (
	SELECT *
	FROM (
		SELECT 1 x, 'foo' as path
		UNION
		SELECT 2 x, 'bar' as path
	) t
	ORDER BY t.x
) t;
-- "baz"	"{foo,bar}"

SELECT '{"foo":{"bar":"baz"}}'::json #>> t.path, t.path
FROM (
	SELECT ARRAY['foo','bar'] as path
) t;
-- "baz"	"{foo,bar}"
```
postgres docs: https://www.postgresql.org/docs/14/functions-json.html#FUNCTIONS-JSON-OP-TABLE
```
json #>> text[] → text
jsonb #>> text[] → text
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -996,6 +1010,14 @@ const typeRules: Dictionary<MatchFn> = {
 			return `TIME(${date})`;
 		}
 	},
+	ToJSON: (args, indent) => {
+		checkMinArgs('ToJSON', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError('ToJSON not supported on: ' + engine);
+		}
+		const column = TextValue(getAbstractSqlQuery(args, 0), indent);
```
Page-, line 1018: This looks like it can be any value, at least according to postgres docs https://www.postgresql.org/docs/14/functions-json.html#FUNCTIONS-JSON-CREATION-TABLE `to_json ( anyelement ) → json`
```suggestion
		const column = AnyValue(getAbstractSqlQuery(args, 0), indent);
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -996,6 +1010,14 @@ const typeRules: Dictionary<MatchFn> = {
...
+		return `to_json(${column})::text`;
```
Page-, line 1019: This should just convert to json
```suggestion
		return `to_json(${column})`;
```
and if necessary the caller can do
```js
['Cast', ['ToJSON', x], 'Text']
```
to cast it back to text

src/AbstractSQLRules2SQL.ts
```diff
@@ -889,6 +892,17 @@ const typeRules: Dictionary<MatchFn> = {
...
+	ExtractJSONPathAsText: (args, indent) => {
+		checkMinArgs('ExtractJSONPathAsText', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError(
+				'ExtractJSONPathAsText not supported on: ' + engine,
+			);
+		}
+		const column = TextValue(getAbstractSqlQuery(args, 0), indent);
+		const path = args[1] ? getAbstractSqlQuery(args, 1)[1] : '';
+		return `${column}::jsonb #>> '{${path}}'`;
```
Page-, line 904: Same thing here with the casting
```suggestion
		return `${column} #>> '{${path}}'`;
```
and then the caller has the option of
```js
['ExtractJSONPathAsText', ['Cast', x, 'JSON'], path]
```
One notable issue is that since the casts look in sbvr-types for the db type https://github.com/balena-io-modules/abstract-sql-compiler/blob/v7.20.2/src/AbstractSQLRules2SQL.ts#L750-L769 there would currently be issues due to the JSON type using TEXT atm: https://github.com/balena-io-modules/sbvr-types/blob/v3.4.8/src/types/json.ts#L4

src/AbstractSQLRules2SQL.ts
```diff
@@ -77,6 +77,8 @@ const UnknownValue: MetaMatchFn = (args, indent) => {
 		case 'Bind':
 		case 'Cast':
 		case 'Coalesce':
+		case 'JSONPath':
```
Page-, line 80: You've got a few (leftover?) `JSONPath` references around the place with no handler for them

src/AbstractSQLCompiler.ts
Page-: It should be a `string[]` based on the postgres docs at least: https://www.postgresql.org/docs/14/functions-json.html#FUNCTIONS-JSON-OP-TABLE
```
json #>> text[] → text
jsonb #>> text[] → text
```

src/AbstractSQLCompiler.ts
Page-: I think we can treat it the same here and rely on postgres complaining

src/AbstractSQLCompiler.ts
```diff
@@ -103,15 +103,18 @@ export type CoalesceNode = [
 	UnknownTypeNodes,
 	...UnknownTypeNodes[],
 ];
+export type ToJSONNode = ['ToJSON', UnknownTypeNodes];
```
Page-, line 106: According to postgres docs at least any element is allowed: https://www.postgresql.org/docs/14/functions-json.html#FUNCTIONS-JSON-CREATION-TABLE `to_json ( anyelement ) → json`
```suggestion
export type ToJSONNode = ['ToJSON', AnyTypeNodes];
```
where `AnyTypeNodes` is basically `Exclude<AbstractSqlType, 'string'>`, ie
```ts
export type AnyTypeNodes =
	| NullNode
	| DateNode
	| BooleanTypeNodes
	| NumberTypeNodes
	| TextTypeNodes
	| UnknownTypeNodes
	| DurationNode
	| SelectQueryNode
	| InsertQueryNode
	| UpdateQueryNode
	| DeleteQueryNode
	| UpsertQueryNode
	| SelectNode
	| ValuesNode
	| InnerJoinNode
	| LeftJoinNode
	| RightJoinNode
	| FullJoinNode
	| CrossJoinNode
	| GroupByNode
	| HavingNode
	| UnknownNode;
export type AbstractSqlType =
	| string
	| AnyTypeNodes;
```
Although I think `string` is only in `AbstractSqlType` because we don't yet have full typing for all the nodes so it's a bit of a catch-all to allow
```ts
export interface AbstractSqlQuery extends Array<AbstractSqlType> {
	0: string;
}
```
to work as a placeholder until we have full typing

joshbwlng: Alright, I'm started to see a pattern. I was struggling with how far I could/should tie things to Postgres docs as this is meant to support MySQL/SQLite as well.

src/AbstractSQLRules2SQL.ts
```diff
@@ -996,6 +1010,14 @@ const typeRules: Dictionary<MatchFn> = {
 			return `TIME(${date})`;
 		}
 	},
+	ToJSON: (args, indent) => {
+		checkMinArgs('ToJSON', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError('ToJSON not supported on: ' + engine);
+		}
+		const column = AnyValue(getAbstractSqlQuery(args, 0), indent);
+		return `to_json(${column})`;
```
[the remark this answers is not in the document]
joshbwlng: @Page- Well, not everywhere :wink: 
https://github.com/balena-io-modules/abstract-sql-compiler/blob/master/src/AbstractSQLRules2SQL.ts#L811

src/AbstractSQLRules2SQL.ts
joshbwlng: Yeah, that was a mistake while doing some debugging. It's now set as a TextValue and I've removed it from the UnknownValue list.

src/AbstractSQLRules2SQL.ts
joshbwlng: My initial opinion is that the SBVR type should use proper JSON at some point, and we could leave this as-is for now (expecting actual JSON). Wouldn't really want a workaround for a workaround.

test/abstract-sql/extract-json-path-as-text.ts
```diff
@@ -0,0 +1,69 @@
...
+						'ExtractJSONPathAsText',
+						['ReferencedField', 'foo', 'bar'],
+						['JSONPath', 't.path', 't.path'],
+					],
+				],
+			],
+		],
+		(result, sqlEquals) => {
+			it('should produce a valid json path text select statement with given targets', () => {
+				sqlEquals(result.query, `SELECT "foo"."bar" #>> t.path, t.path`);
```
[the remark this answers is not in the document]
joshbwlng: I only included this test as it was an example you provided of non-literal array cases when using `#>>`. I'll go ahead and replace it with something more realistic, even though the main point of the test itself was to confirm that providing multiple strings to `JSONPath` results in a generated query that includes all and it works as expected.

test/abstract-sql/extract-json-path-as-text.ts
```diff
@@ -0,0 +1,69 @@
...
+		[
+			'SelectQuery',
+			[
+				'Select',
+				[
+					[
+						'ExtractJSONPathAsText',
+						['ReferencedField', 'foo', 'bar'],
+						['JSONPath', `'{buz,baz}'`],
+					],
```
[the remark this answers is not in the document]
joshbwlng: Are you saying there's a case in which tests would pass, but we'd get complaints in production?

src/AbstractSQLCompiler.ts
```diff
@@ -125,11 +128,18 @@ export type ReplaceNode = [
 	TextTypeNodes,
 	TextTypeNodes,
 ];
+export type JSONPathNode = ['JSONPath', ...TextTypeNodes[]];
+export type ExtractJSONPathAsTextNode = [
+	'ExtractJSONPathAsText',
+	JSONNode,
+	JSONPathNode?,
+];
```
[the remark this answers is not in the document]
joshbwlng: Ok, that makes sense. I was following the pattern found in `ReferencedField` for providing multiple strings within a single node. However this makes much more sense, by separating it out into multiple `TextNode[]`s, the caller can provide different types of text nodes and not just explicit strings.

test/abstract-sql/extract-json-path-as-text.ts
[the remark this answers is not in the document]
joshbwlng: Nevertheless, your suggestion to use `ARRAY[]` makes much more sense than my literal string approach, thanks!

src/AbstractSQLOptimiser.ts
```diff
@@ -672,6 +677,24 @@ const typeRules: Dictionary<MatchFn> = {
...
+		checkMinArgs('ExtractJSONPathAsText', args, 1);
+		const column = TextValue(getAbstractSqlQuery(args, 0));
+		const path = args.slice(1).map((arg) => {
+			if (!isAbstractSqlQuery(arg)) {
+				throw new SyntaxError(
+					`Expected AbstractSqlQuery array but got ${typeof arg}`,
+				);
+			}
+			return ArrayValue(arg);
+		});
```
Page-, line 690: Since there's only one path arg according to the types then this can be:
```suggestion
		const path = ArrayValue(getAbstractSqlQuery(args, 1));
```

src/AbstractSQLOptimiser.ts
```diff
@@ -672,6 +677,24 @@ const typeRules: Dictionary<MatchFn> = {
...
+		return ['ExtractJSONPathAsText', column, ...path];
+	},
+	TextArray: (args) => {
+		// Allow for populated and empty arrays
+		return ['TextArray', ...args];
```
Page-, line 695: To enforce each element of the array is a text value:
```suggestion
		return ['TextArray', ...args.map(TextValue)];
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -889,6 +910,19 @@ const typeRules: Dictionary<MatchFn> = {
...
 		return `REPLACE(${str}, ${find}, ${replacement})`;
 	},
+	ExtractJSONPathAsText: (args, indent) => {
+		checkMinArgs('ExtractJSONPathAsText', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError(
+				'ExtractJSONPathAsText not supported on: ' + engine,
+			);
+		}
+		const column = TextValue(getAbstractSqlQuery(args, 0), indent);
```
Page-, line 920: This should be a `JSONValue` (`isJSONValue` will need updating to include `ToJSON` and be wrapped in a `MatchValue`), trying JSON operators on a text type won't work, also renamed as `json` rather than `column` since it doesn't need to be a column and can just be any JSON type
```suggestion
		const json = JSONValue(getAbstractSqlQuery(args, 0), indent);
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -889,6 +910,19 @@ const typeRules: Dictionary<MatchFn> = {
...
+		const path = args[1]
+			? TextValue(getAbstractSqlQuery(args, 1), indent)
+			: `CAST(ARRAY[] as TEXT[])`;
```
Page-, line 923: The typings say that the path must always be specified:
```suggestion
		const path = TextValue(getAbstractSqlQuery(args, 1), indent);
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -996,6 +1030,14 @@ const typeRules: Dictionary<MatchFn> = {
 			return `TIME(${date})`;
 		}
 	},
+	ToJSON: (args, indent) => {
+		checkMinArgs('ToJSON', args, 1);
+		if (engine !== Engines.postgres) {
+			throw new SyntaxError('ToJSON not supported on: ' + engine);
+		}
+		const column = AnyValue(getAbstractSqlQuery(args, 0), indent);
```
Page-, line 1038: This doesn't need to be a column specifically so it's better to use a more generic name to avoid confusion:
```suggestion
		const value = AnyValue(getAbstractSqlQuery(args, 0), indent);
```

PR #151, opened 2022-11-16
Fix date fields used inside arithmetic expressions

fisehara:
isNumericValue is evaluated in arithmetic expressions if operands in expression can be used as numericValues.

In SQL statements arithmetic expressions like add/sub are supported for date values.
DateArithmetic nodes need input dependent typing.

Change-type: minor
Signed-off-by: fisehara <harald@balena.io>

### DIFF
--- a/src/AbstractSQLOptimiser.ts
+++ b/src/AbstractSQLOptimiser.ts
@@ -385,6 +385,17 @@ const JoinMatch =
 		}
 	};
 
+const AddDateMatcher = tryMatches(
+	matchArgs('AddDateDuration', DateValue, DurationValue),
+	matchArgs('AddDateNumber', DateValue, NumericValue),
+);
+
+const SubtractDateMatcher = tryMatches(
+	matchArgs('SubtractDateDate', DateValue, DateValue),
+	matchArgs('SubtractDateDuration', DateValue, DurationValue),
+	matchArgs('SubtractDateNumber', DateValue, NumericValue),
+);
+
 const typeRules: Dictionary<MatchFn> = {
 	UnionQuery: (args) => {
 		checkMinArgs('UnionQuery', args, 2);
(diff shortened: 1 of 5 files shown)


src/AbstractSQLRules2SQL.ts
```diff
@@ -142,7 +142,9 @@ export const isNumericValue = (
 		type === 'Ceiling' ||
 		type === 'Count' ||
 		type === 'Average' ||
-		type === 'Sum'
+		type === 'Sum' ||
+		isDateValue(type) ||
+		isDurationValue(type)
```
Page-, line 147: These are not numeric values and it's actually more accurate to say that you can do `number + number` or `date + number` but not `date + date` but you can do `date - date`, see https://www.postgresql.org/docs/14/functions-datetime.html for a list of the operators for date/time types in postgres

test/abstract-sql/aggregate.ts
```diff
@@ -486,3 +486,117 @@ describe('ToTime', () => {
...
+	);
+});
+
+describe('Add now timestamp to now timestamp', () => {
+	test(
+		['SelectQuery', ['Select', [['Add', ['Now'], ['Now']]]]],
+		(result, sqlEquals) => {
+			it('should produce a valid Subtract statement', () => {
+				console.log(`result:${JSON.stringify(result, null, 2)}`);
+				sqlEquals(result.query, 'SELECT CURRENT_TIMESTAMP + CURRENT_TIMESTAMP');
```
Page-, line 526: This is invalid
```
ERROR:  operator does not exist: timestamp with time zone + timestamp with time zone
LINE 1: SELECT CURRENT_TIMESTAMP + CURRENT_TIMESTAMP

**Review by Page- — CHANGES_REQUESTED** [review]

File: src/AbstractSQLOptimiser.ts [hunk 3]
```diff
@@ -627,8 +627,47 @@ const typeRules: Dictionary<MatchFn> = {
...
+	AddDate: tryMatches(
+		// checks every matcher and will proceed with the first one satisfying the TypeMatch
+		// Fall through helper matcher to throw an error when the operand types don't match.
+		matchArgs('AddDate', DateValue, DurationValue),
+		matchArgs('AddDate', DateValue, NumericValue),
+		Helper<OptimisationMatchFn>((args) => {
+			throw new SyntaxError(
+				`"AddDate" expects (DateValue, DurationValue | NumericValue) but given ${args}`,
+			);
+		}),
```
Page-, line 669: The helper wrapper should be unnecessary?
```suggestion
		(args) => {
			throw new SyntaxError(
				`"AddDate" expects (DateValue, DurationValue | NumericValue) but given ${args}`,
			);
		},
```

src/AbstractSQLRules2SQL.ts
```diff
@@ -831,6 +833,31 @@ const typeRules: Dictionary<MatchFn> = {
...
+		}
+	},
+	SubtractDate: (args, indent) => {
+		const a = AnyValue(getAbstractSqlQuery(args, 0), indent);
+		const b = AnyValue(getAbstractSqlQuery(args, 1), indent);
+
+		if (engine === Engines.postgres) {
+			return `${a} - ${b}`;
+		} else if (engine === Engines.mysql) {
+			return `DATE_SUB(${a}, ${b})`;
```
Page-, line 856: If `a` and `b` are both dates then this would need to be `DATEDIFF`

src/AbstractSQLRules2SQL.ts
```diff
@@ -443,6 +453,75 @@ export const checkMinArgs = (matchName: string, args: any[], num: number) => {
 	}
 };
 
+const AddDateNumber: MatchFn = (args, indent) => {
+	checkArgs('AddDateNumber', args, 2);
+	const a = DateValue(getAbstractSqlQuery(args, 0), indent);
+	const b = AnyValue(getAbstractSqlQuery(args, 1), indent);
```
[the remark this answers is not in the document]
fisehara: Oh right, thanks!

fisehara:
@balena-ci I self-certify!

PR #163, opened 2022-11-17
Dev: Replace husky with simple-git-hooks

joshbwlng:
Change-type: patch

---

`husky` v5 and later caused a number of issues that caused the Jellyfish team to move to `simple-git-hooks` quite a while back. I don't remember the exact issue, but I believe it had something to do with hook installation. `simple-git-hooks` is quite simple as the name suggests and solves the problem without the extra headache.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,17 +32,15 @@
     "@types/mocha": "^10.0.0",
     "chai": "^4.3.7",
     "common-tags": "^1.8.2",
-    "husky": "^4.3.8",
     "lint-staged": "^13.0.3",
     "mocha": "^10.1.0",
     "require-npm4-to-publish": "^1.0.0",
+    "simple-git-hooks": "^2.8.1",
     "ts-node": "^10.9.1",
     "typescript": "^4.9.3"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
+  "simple-git-hooks": {
+    "pre-commit": "npx lint-staged"
   },
   "lint-staged": {
     "*.js": [


Page- (commented):
We use husky everywhere else so I'd like to see what the benefit is with switching? And if it makes sense then we should switch all our husky usage over

joshbwlng:
@Page- I just noticed that husky seems to be v4 in most places, `balena-api` has v8 but needs us to execute `husky.install()` like in https://github.com/balena-io/balena-api/blob/master/package.json#L30. Pretty sure we don't have to do this type of stuff with `simple-git-hooks`. I'll look at what happens when we replace husky with simple-git-hooks in the `balena-api` repo. And what happens when we update this repo from husky v4 to v8.

joshbwlng:
Ah, renovate went ahead and bumped husky to v8 last night. Doubt its working as expected as its not executing `install()` etc. I'll check it out.

joshbwlng:
And I remember now, I didn't like the need to execute `husky.install()` in `prepare` etc scripts and wanted something that was less involved and just worked. And that was `simple-git-hooks`. I prefer having a couple lines of config instead of needing to execute a command. I would seriously consider moving everything to `simple-git-hooks` unless there's something that makes us need `husky`. Technically though we could continue to use `husky`, and update all repos to v8, if we don't mind executing `husky.install()` everywhere. :tired_face:

fisehara:
@joshbwlng @Page- husky 8.0.2 has now a way to call 'husky install' instead of invoking it by `require('husky').install()` 

Should we follow this approach on updating everything to husky8?

Page-:
@fisehara the way to do it is https://github.com/balena-io/open-balena-api/blob/v5.3.6/package.json#L32
```
    "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\""
```
so that it also works when husky is not installed (eg `npm ci --production`) and is cross platform.

@joshbwlng simple-git-hooks will automatically install in the first place but not automatically update hooks which means it could end up as a bit of a headache later

joshbwlng:
> husky 8.0.2 has now a way to call 'husky install' instead of invoking it by `require('husky').install()`
> 
> Should we follow this approach on updating everything to husky8?

@fisehara Are you perhaps referring to `npx husky install`? I also saw `npm pkg set scripts.prepare="husky install"` in the husky README, but that just results in what Page mentioned above.

@Page- Could we not simplify that node call to something like `npx --no husky install`?

PR #175, opened 2022-11-23
Add binds for affected IDs in compiled rules

Ereski:
After every write, pine has to rerun all rules from the model to ensure consistency. These rules run over the entire database, sometimes causing some queries to run for too long against what should be a simple write.

This commit adds a mechanism to help with this issue by narrowing the set of rows that each rule should touch to those rows that were actually changed.

Implementing this mechanism safely is doable and not necessarily complex code-wise, but requires a deep modifications from the current architecture. This commit adds a restricted form instead where we only narrow the rows of the root table that were changed. If any other table was changed then narrowing is a no op.

It can be proved that this is always safe as long as the root table is selected from only once and the rule is positive ("It is necessary that each ...").

The implementation here adds a single binding into the rule's SQL query which can be bound by pine for each rule where an opportunity to use this optimization arises.

The implementation itself is simple: count how many times the root table is selected from and if it is selected from exactly one, then add a narrowing constraint in the form of:

$1 = '{}' OR
<root table>.id = ANY(CAST($1 AS INTEGER[]))

Where $1 will be bound to either '{}', which disables narrowing, or to a list of IDs that were affected by the write.

This initial implementation can be extended in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "lodash": "^4.17.21"
   },
   "devDependencies": {
-    "@balena/lf-to-abstract-sql": "^4.6.0",
+    "@balena/lf-to-abstract-sql": "^5.0.0",
     "@balena/lint": "^6.2.1",
     "@balena/odata-parser": "^2.4.2",
     "@balena/odata-to-abstract-sql": "^5.8.0",
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -21,6 +21,7 @@ import {
 	ReferencedFields,
 	RuleReferencedFields,
 	ModifiedFields,
+	insertAffectedIdsBinds,
 } from './referenced-fields';
 
 export type { ReferencedFields, RuleReferencedFields, ModifiedFields };
@@ -187,24 +188,21 @@ export type TextTypeNodes =
 	| ExtractJSONPathAsTextNode
 	| UnknownTypeNodes;
 
-export type SelectQueryNode = [
-	'SelectQuery',
-	...Array<
-		| SelectNode
-		| FromNode
-		| InnerJoinNode
-		| LeftJoinNode
-		| RightJoinNode
-		| FullJoinNode
-		| CrossJoinNode
-		| WhereNode
-		| GroupByNode
-		| HavingNode
-		| OrderByNode
-		| LimitNode
-		| OffsetNode
-	>,
-];
+export type SelectQueryStatementNode =
+	| SelectNode
+	| FromNode
+	| InnerJoinNode
+	| LeftJoinNode
+	| RightJoinNode
+	| FullJoinNode
+	| CrossJoinNode
+	| WhereNode
+	| GroupByNode
+	| HavingNode
+	| OrderByNode
+	| LimitNode
+	| OffsetNode;
+export type SelectQueryNode = ['SelectQuery', ...SelectQueryStatementNode[]];
 export type UnionQueryNode = [
 	'UnionQuery',
 	// tslint:disable-next-line:array-type typescript fails on a circular reference when `Array<T>` form
@@ -238,7 +236,7 @@ export type FromTypeNodes =
 	| FromTypeNode[keyof FromTypeNode]
 	| AliasNode<FromTypeNode[keyof FromTypeNode]>;
 
-type AliasableFromTypeNodes = FromTypeNodes | AliasNode<FromTypeNodes>;
+export type AliasableFromTypeNodes = FromTypeNodes | AliasNode<FromTypeNodes>;
 
 export type SelectNode = ['Select', AbstractSqlType[]];
 export type FromNode = ['From', AliasableFromTypeNodes];
@@ -396,6 +394,17 @@ export interface AbstractSqlModel {
 		body: string;
 		language: 'plpgsql';
 	}>;
+	lfInfo: {
+		rules: {
+			[key: string]: LfRuleInfo;
+		};
+	};
+}
+export interface LfRuleInfo {
+	root: {
+		table: string;
+		alias: string;
+	};
 }
 export interface SqlModel {
 	synonyms: {
@@ -503,6 +512,8 @@ export const isSelectQueryNode = (n: AbstractSqlType): n is SelectQueryNode =>
 	n[0] === 'SelectQuery';
 export const isSelectNode = (n: AbstractSqlType): n is SelectNode =>
 	n[0] === 'Select';
+export const isWhereNode = (n: AbstractSqlType): n is WhereNode =>
+	n[0] === 'Where';
 
 /**
  *
@@ -873,6 +884,7 @@ CREATE TABLE ${ifNotExistsStr}"${table.name}" (
 			if (typeof ruleSE !== 'string') {
 				throw new Error('Invalid structured English');
 			}
+			insertAffectedIdsBinds(ruleBody, abstractSqlModel.lfInfo.rules[ruleSE]);
 			const { query: ruleSQL, bindings: ruleBindings } = compileRule(
 				ruleBody,
 				engine,
--- a/src/referenced-fields.ts
+++ b/src/referenced-fields.ts
@@ -2,16 +2,55 @@ import * as _ from 'lodash';
 import {
 	AbstractSqlQuery,
 	AbstractSqlType,
+	AddDateDurationNode,
+	AddDateNumberNode,
+	AliasNode,
+	AndNode,
+	AnyNode,
+	AverageNode,
+	CountNode,
+	CrossJoinNode,
+	DateTruncNode,
 	EngineInstance,
+	EqualsNode,
+	ExistsNode,
 	FieldsNode,
+	FromNode,
+	FromTypeNodes,
+	FullJoinNode,
+	GreaterThanNode,
+	GreaterThanOrEqualNode,
+	HavingNode,
+	InnerJoinNode,
+	InNode,
 	isAliasNode,
 	isFromNode,
 	isSelectNode,
 	isSelectQueryNode,
 	isTableNode,
+	isWhereNode,
+	LeftJoinNode,
+	LessThanNode,
+	LessThanOrEqualNode,
+	LfRuleInfo,
+	NotEqualsNode,
+	NotExistsNode,
+	NotNode,
+	OrNode,
+	RightJoinNode,
 	SelectNode,
+	SelectQueryNode,
+	SubtractDateDateNode,
+	SubtractDateDurationNode,
+	SubtractDateNumberNode,
+	SumNode,
+	TableNode,
+	ToJSONNode,
+	UnionQueryNode,
+	WhereNode,
 } from './AbstractSQLCompiler';
 import { AbstractSQLOptimiser } from './AbstractSQLOptimiser';
+import { isAbstractSqlQuery } from './AbstractSQLRules2SQL';
 
 export interface ReferencedFields {
 	[alias: string]: string[];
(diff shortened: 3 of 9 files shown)


src/referenced-fields.ts
```diff
@@ -294,3 +313,271 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+// - Ban on aggregates is not necessary but solving it requires additional
+//   complexity.
+// - We only care about and recognize `COUNT` aggregates and `GROUP BY`
+//   statements right now.
+// - We assume the ID column is named "id".
+// - There are other techniques that cover cases that this simple narrowing
+//   does not. Currently that are two with their own kinks to work out:
+//   ID propagation and COUNT splitting.
+export const addAffectedIdsBinds = (
+	abstractSql: AbstractSqlQuery,
```
fisehara, line 517: Please add some comment that the input data is manipulated in situ / in place.

test/sbvr/pilots.js
```diff
@@ -160,6 +160,8 @@ SELECT (
 		FROM "pilot-can fly-plane" AS "pilot.0-can fly-plane.1"
 		WHERE "pilot.0-can fly-plane.1"."pilot" = "pilot.0"."id"
 	)
+	AND ($1 = '{}'
+	OR "pilot.0"."id" = ANY(CAST($1 AS INTEGER[])))
```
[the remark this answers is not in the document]
fisehara:
```
	AND ($1 = '{}'
	OR "pilot.0"."id" = ANY(CAST($1 AS INTEGER[])))
```
I'm very sorry but I don't understand what this is doing.

fisehara (commented):
Tried, but don't understand fully whats happening

Ereski:
@fisehara The context for all of this is here: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/Narrowing.20rows.20when.20running.20rules/near/321176832. Caveat emptor, huge thread and something like 99% percent of it is not applicable anymore. But this PR is still relevant so I'll try to summarize in zulip what's going on here

test/sbvr/pilots.js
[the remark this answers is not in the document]
Ereski: Those IDs are passed as strings convertible to arrays (cause of limitations on how we can pass values to binds...). So an empty array `'{}'` means that we don't narrow this table and when it's not an empty array we narrow by the ids bound

src/AbstractSQLCompiler.ts
```diff
@@ -146,13 +147,15 @@ export type CoalesceNode = [
 	...UnknownTypeNodes[],
 ];
 export type ToJSONNode = ['ToJSON', AnyTypeNodes];
+export type AnyNode = ['Any', UnknownTypeNodes];
```
Page-, line 150: If you can split the addition of the `Any` node out to a separate PR that'd be very helpful to keep PRs as small as possible and much easier/quicker to review, meaning all the PRs can get merged sooner and with less issues around rebase conflicts

src/referenced-fields.ts
```diff
@@ -294,3 +313,271 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+		case 'LessThanOrEqual':
+		case 'IsDistinctFrom':
+		case 'IsNotDistinctFrom':
+			const equalsNode = constraint as EqualsNode;
+			const left = equalsNode[1];
+			const right = equalsNode[2];
+			if (!isAbstractSqlQuery(left) || !isAbstractSqlQuery(right)) {
+				throw new Error(
+					'cannot introspect into the string form of AbstractSqlQuery',
+				);
```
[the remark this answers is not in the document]
Ereski: Not at all.  This checks for the string form of `AbstractSqlQuery` (https://github.com/balena-io-modules/abstract-sql-compiler/blob/master/src/AbstractSQLCompiler.ts#L298), not string literals (IIRC, https://github.com/balena-io-modules/abstract-sql-compiler/blob/master/src/AbstractSQLCompiler.ts#L158)

src/referenced-fields.ts
```diff
@@ -2,16 +2,35 @@ import * as _ from 'lodash';
...
 	SelectNode,
+	SelectQueryNode,
+	SelectQueryStatementNode,
+	TableNode,
+	WhereNode,
 } from './AbstractSQLCompiler';
 import { AbstractSQLOptimiser } from './AbstractSQLOptimiser';
+import { isAbstractSqlQuery } from './AbstractSQLRules2SQL';
+
+const SELECT_QUERY_TYPES = ['SelectQuery', 'UnionQuery'];
```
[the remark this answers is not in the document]
Ereski: Sure, just didn't find a reason to after testing everything in the tree up to the api

src/referenced-fields.ts
```diff
@@ -294,3 +313,271 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+			return null;
+	}
+};
+
+const findBindCandidates = (
+	statement: AbstractSqlQuery,
+	candidates: BindCandidate[],
+) => {
+	switch (statement[0]) {
+		case 'SelectQuery':
```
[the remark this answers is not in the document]
Ereski: Same as above

src/AbstractSQLCompiler.ts
```diff
@@ -867,6 +877,7 @@ CREATE TABLE ${ifNotExistsStr}"${table.name}" (
 			if (typeof ruleSE !== 'string') {
 				throw new Error('Invalid structured English');
 			}
+			addAffectedIdsBinds(ruleBody, abstractSqlModel.lfInfo.rules[ruleSE]);
```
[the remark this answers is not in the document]
Ereski: It depends but I would rather not. This function shouldn't throw for any rule unless it contains unsupported nodes (like `UnionQuery`) or legacy structures (like the string form of `AbstractSqlQuery`. This code runs at compile time so it's wouldn't cause random errors at runtime and I think it's best to fix/implement whatever is is missing than skip it.

src/referenced-fields.ts
Ereski: I would think that the name itself `addAffectedIdsBinds` would be enough to show that, specially because this function doesn't return anything. Maybe `decorate` is better than `add` to convey this?

src/AbstractSQLCompiler.ts
Ereski: Will do. Commit history is a mess rn

src/referenced-fields.ts
```diff
@@ -294,3 +313,271 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+			const equalsNode = constraint as EqualsNode;
+			const left = equalsNode[1];
+			const right = equalsNode[2];
+			if (!isAbstractSqlQuery(left) || !isAbstractSqlQuery(right)) {
+				throw new Error(
+					'cannot introspect into the string form of AbstractSqlQuery',
+				);
+			}
+			if (SELECT_QUERY_TYPES.includes(left[0])) {
+				findBindCandidates(left, candidates);
```
Page-, line 351: I think it would make sense for `findBindCandidates` to do:
```ts
if (!SELECT_QUERY_TYPES.includes(statement[0])) {
	return;
}
```
at the start rather than having to guard every call into it with this duplicated check

src/referenced-fields.ts
```diff
@@ -294,3 +313,271 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+				findBindCandidates(inner, candidates);
+			}
+			break;
+		case 'Not':
+			const notNode = constraint as NotNode;
+			findTablesInConstraint(notNode[1], candidates);
+			break;
+		case 'And':
+		case 'Or':
+			const andNode = constraint as AndNode;
```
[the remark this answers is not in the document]
Page-: If you use
```suggestion
			const andNode = constraint as AndNode | OrNode;
```
does it still all type-check correctly?

src/referenced-fields.ts
```diff
@@ -294,3 +313,271 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+			for (const arg of andNode.slice(1)) {
+				// TODO: type checking is going awry here for some reason
+				findTablesInConstraint(arg as any, candidates);
+			}
```
Page-, line 394: It might work if you did something like
```suggestion
			const [ , ...args] = andNode;
			for (const arg of args) {
				findTablesInConstraint(arg, candidates);
			}
```
as I expect the issue is that `.slice` returns `(typeof andNode)[number]` and doesn't play nice with tuples

Ereski:
@Page- sorry I pushed at the same time as your review. I'll try to find the lines you quoted

Ereski:
Your suggestion:

```javascript
const andNode = constraint as AndNode | OrNode;
```

Typechecks. I can't reply directly to that but I'll include this change

src/referenced-fields.ts
Ereski: I just fixed this. It was missing an `!isAbstractSqlQuery` check which does the coercion.

src/referenced-fields.ts
Page-: What I mean is that instead of having
```ts
if (SELECT_QUERY_TYPES.includes(biOpLeft[0])) {
	findBindCandidates(biOpLeft, candidates);
}
if (SELECT_QUERY_TYPES.includes(biOpRight[0])) {
	findBindCandidates(biOpRight, candidates);
}
....
const findBindCandidates = (statement, candidates) => {
	...
}
```
that you have
```ts
findBindCandidates(biOpLeft, candidates);
findBindCandidates(biOpRight, candidates);
....
const findBindCandidates = (statement, candidates) => {
	if (!SELECT_QUERY_TYPES.includes(statement[0])) {
		return;
	}
	...
}
```
in order to reduce the duplication and remove the onus on the callsite by moving the parameter checking into the function itself

Ereski: Ah that makes sense :+1:

Ereski:
@Page- added a summary ^^^^^

What the code does is simpler than it looks

src/AbstractSQLCompiler.ts
```diff
@@ -576,6 +582,8 @@ const compileSchema = (
 	engine: Engines,
 	ifNotExists: boolean,
 ): SqlModel => {
+	abstractSqlModel = optimizeSchema(abstractSqlModel, false);
```
Page-, line 585: If you can chuck this in a separate PR that'd be great

src/referenced-fields.ts
```diff
@@ -294,3 +322,302 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+	const selectQueryNode = rootTableCandidate.scope as SelectQueryNode;
+	let whereNode: WhereNode | null = null;
+	for (const statement of selectQueryNode.slice(1)) {
+		if (statement[0] === 'Where') {
+			whereNode = statement;
+
+			break;
+		}
+	}
+	if (!whereNode) {
```
Page-, line 601: It's best to explicitly compare against null when that's the intention as it avoids any potential confusion as to the intent (whether the intent is `== null` or `== null || === 0 || === '' || ...`
```suggestion
	if (whereNode == null) {
```

src/referenced-fields.ts
```diff
@@ -294,3 +322,302 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+
+	const selectQueryNode = rootTableCandidate.scope as SelectQueryNode;
+	let whereNode: WhereNode | null = null;
+	for (const statement of selectQueryNode.slice(1)) {
+		if (statement[0] === 'Where') {
+			whereNode = statement;
+
+			break;
+		}
+	}
```
Page-, line 600:
```suggestion
	const whereNode = selectQueryNode.slice(1).find(isWhereNode);
```
where `isWhereNode` is
```ts
export const isWhereNode = (n: AbstractSqlType): n is WhereNode =>
	n[0] === 'Where';

File: src/referenced-fields.ts [hunk 14]
```diff
@@ -294,3 +322,302 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+	// Count the number of references for each database table. Multiple
+	// candidates referencing the same table will be removed to conservatively
+	// avoid issues
+	const seenTableNames: { [key: string]: number } = {};
+	for (const candidate of candidates) {
+		if (candidate.tableName in seenTableNames) {
+			seenTableNames[candidate.tableName] += 1;
+		} else {
+			seenTableNames[candidate.tableName] = 1;
+		}
```
Page-, line 571: I think this is slightly simpler using the `??=` operator to initialize to 0
```suggestion
		seenTableNames[candidate.tableName] ??= 0;
		seenTableNames[candidate.tableName] += 1;
```

src/referenced-fields.ts
```diff
@@ -294,3 +322,302 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+			findTablesInConstraint(whereNode[1], candidates);
+
+			return null;
+
+		default:
+			return null;
+	}
+};
+
+const findBindCandidates = (
```
Page-, line 501: Some jsdoc detailing the intention of this function would be useful, I initially thought it would be looking for bind vars but it looks like it's actually looking for queries into which to add bind vars. And the same thing for the other functions - I have a feeling a bunch of them can be named more clearly and possibly collapsed but it's hard to say for sure right now as I don't fully understand all of the interactions

src/AbstractSQLCompiler.ts
Ereski: https://github.com/balena-io-modules/abstract-sql-compiler/pull/190

package.json
```diff
@@ -20,7 +20,7 @@
     "lodash": "^4.17.21"
   },
   "devDependencies": {
-    "@balena/lf-to-abstract-sql": "^4.6.0",
+    "@balena/lf-to-abstract-sql": "5.0.0-build-expand-lfinfo-cd6429c29d11d986d13b2f3a1c7090310e3e4d03-1",
```
Page-, line 23: This should point to a real version before unmarking as draft (to avoid accidentally merging whilst pointing at a prerelease version)

src/referenced-fields.ts
```diff
@@ -294,3 +332,256 @@ export const getModifiedFields: EngineInstance['getModifiedFields'] = (
...
+	}
+
+	const selectQueryNode = abstractSql[1] as SelectQueryNode;
+	const whereStatement = selectQueryNode.slice(1).find((x) => x[0] === 'Where');
+	if (whereStatement === undefined) {
+		selectQueryNode.push(['Where', narrowing]);
+	} else {
+		const whereNode = whereStatement as WhereNode;
+		whereNode[1] = ['And', whereNode[1], narrowing];
+	}
```
Page-, line 586: I assume there's an `isWhereNode` (if not it makes sense to create imo):
```suggestion
	const whereNode = selectQueryNode.slice(1).find(isWhereNode);
	if (whereNode === undefined) {
		selectQueryNode.push(['Where', narrowing]);
	} else {
		whereNode[1] = ['And', whereNode[1], narrowing];
	}
```

Page- (commented):
Looks sane, just minor things (apart from the module version being necessary)

package.json
```diff
@@ -20,7 +20,7 @@
     "lodash": "^4.17.21"
   },
   "devDependencies": {
-    "@balena/lf-to-abstract-sql": "^4.6.0",
+    "@balena/lf-to-abstract-sql": "5.0.0",
```
Page-, line 23:
```suggestion
    "@balena/lf-to-abstract-sql": "^5.0.0",
```

Ereski: fixed

Page- (approved):

PR #177, opened 2022-11-29
Proposal: v9

Page-:


test/sbvr/pilots.js
```diff
@@ -169,7 +169,7 @@ SELECT NOT EXISTS (
 SELECT NOT EXISTS (
 	SELECT 1
 	FROM "pilot" AS "pilot.0"
-	WHERE "pilot.0"."is experienced" = 1
+	WHERE "pilot.0"."is experienced" = TRUE
```
thgreasi, line 172: I guess changing to real booleans allows us to do an extra optimization for PG in a follow-up minor PR that will convert this to
```
WHERE "pilot.0"."is experienced"
```
right? Extra 4 bytes sql statement size savings :tada:

package.json
```diff
@@ -16,11 +16,11 @@
   "repository": "https://github.com/balena-io-modules/abstract-sql-compiler.git",
```
thgreasi, line 16: Should we also bump node and add an engines field?

thgreasi (commented):
So far lgtm :+1:

test/sbvr/pilots.js
Page-: Yeah that should be possible in future

package.json
```diff
@@ -57,6 +57,9 @@
       "test/**/*.js"
     ]
   },
+  "engines": {
+    "node": ">=16.13.0"
+  },
```
fisehara, line 62: Drop:
```
"prepublish": "require-npm4-to-publish",
...
"require-npm4-to-publish": "^1.0.0",
```


```suggestion
  "engines": {
    "node": ">=16.13.0",
    "npm": ">=8.1.0"
  },
```

fisehara (commented):
LGTM + NIT

fisehara (commented):
LGTM but please let @thgreasi run through it also.
In addition, I'm not on top of all necessary or pending major changes. But this shouldn't block this major release.

thgreasi (approved):
:tada:

PR #180, opened 2022-11-30
Fix: move @balena/odata-to-abstract-sql to dev dependencies

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #182, opened 2022-12-08
Add support for rules optimizations when using COUNT(*) = 0

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

Issue #184, opened 2023-01-23
Multiply operator should correctly wrap arguments in brackets


JSReds:
As the title say, `Multiply` operator should correctly wrap arguments. 

e.g.:

```
'Multiply',
[
	'Multiply',
	['ReferencedField', 'credit bundle', 'unit cost'],
	['ReferencedField', 'credit bundle', 'original quantity'],
],
[
	'Subtract',
	['Number', 1],
	[
		'ReferencedField',
		'credit bundle',
		'discount percentage',
	],
]
``` 

should return  : ("credit bundle"."unit cost" * "credit bundle"."original quantity") * (1 - "credit bundle"."discount percentage")
but returns: "credit bundle"."unit cost" * "credit bundle"."original quantity" * 1 - "credit bundle"."discount percentage"

e.g. code : https://github.com/balena-io/balena-api/pull/4037/files#r1083707092

PR #185, opened 2023-01-23
Respect math operator precedence

thgreasi:
Resolves: #184
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/345890-loop.2Fbalena-io/topic/Balena.20Pricing.202.2E0
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/absract-sql-compiler.20math.20operator.20precedence

### DIFF
--- a/src/AbstractSQLRules2SQL.ts
+++ b/src/AbstractSQLRules2SQL.ts
@@ -378,11 +378,61 @@ const mathOps = {
 	Multiply: '*',
 	Divide: '/',
 };
+
+const mathOpsPrecedence = (() => {
+	const operatorsByPrecedence = [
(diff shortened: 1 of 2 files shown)


Page- (commented):
@**thgreasi** I think it'd be simpler to go with brackets because then it also avoids any other weird interactions that could happen where the math op is inside of something else where it does need to have brackets, as an example that might break which you can try:
```
['Cast', ['Add', ['Number', 1], ['Number', 2]], 'Text'];

thgreasi:
Closing in favor of the simpler #186

PR #186, opened 2023-01-23
Wrap composite math operations with parenthesis

thgreasi:
Resolves: #184
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/absract-sql-compiler.20math.20operator.20precedence

### DIFF
--- a/src/AbstractSQLRules2SQL.ts
+++ b/src/AbstractSQLRules2SQL.ts
@@ -379,15 +379,25 @@ const mathOps = {
 	Divide: '/',
 };
 
+const mathOperatorNodeTypes = new Set([
+	...Object.keys(mathOps),
+	'AddDateDuration',
+	'AddDateNumber',
+	'SubtractDateDate',
+	'SubtractDateDuration',
+	'SubtractDateNumber',
+]);
+
 const mathOpValue = (
+	valueMatchFn: MetaMatchFn,
 	args: AbstractSqlType[],
 	index: number,
 	indent: string,
 ) => {
 	const operandAbstractSql = getAbstractSqlQuery(args, index);
-	const numericValue = NumericValue(operandAbstractSql, indent);
+	const numericValue = valueMatchFn(operandAbstractSql, indent);
 	const [childNodeType] = operandAbstractSql;
-	if (childNodeType in mathOps) {
+	if (mathOperatorNodeTypes.has(childNodeType)) {
 		return `(${numericValue})`;
 	}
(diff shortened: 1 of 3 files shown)


test/abstract-sql/dates.ts
```diff
@@ -154,3 +154,119 @@ describe('ToTime', () => {
...
+	test(
+		[
+			'SelectQuery',
+			[
+				'Select',
+				[
+					[
+						'Multiply',
+						['SubtractDateDate', ['Now'], ['Now']],
+						['Subtract', ['Number', 4], ['Number', 5]],
```
fisehara, line 258: @thgreasi in this case the date subtraction need to be in parentheses.

src/AbstractSQLRules2SQL.ts
```diff
@@ -452,7 +467,7 @@ export const checkMinArgs = (matchName: string, args: any[], num: number) => {
 const AddDateNumber: MatchFn = (args, indent) => {
 	checkArgs('AddDateNumber', args, 2);
 	const a = DateValue(getAbstractSqlQuery(args, 0), indent);
-	const b = NumericValue(getAbstractSqlQuery(args, 1), indent);
+	const b = mathOpValue(args, 1, indent);
```
fisehara, line 470: Only checking the right site operand is not sufficient. As postgres date functions explicit define the different outcome from operations bound to the left and right operant of date arithmetics:
https://www.postgresql.org/docs/current/functions-datetime.html

src/AbstractSQLRules2SQL.ts
```diff
@@ -494,7 +509,7 @@ const SubtractDateDuration: MatchFn = (args, indent) => {
 const SubtractDateNumber: MatchFn = (args, indent) => {
 	checkArgs('SubtractDateNumber', args, 2);
 	const a = DateValue(getAbstractSqlQuery(args, 0), indent);
-	const b = NumericValue(getAbstractSqlQuery(args, 1), indent);
+	const b = mathOpValue(args, 1, indent);
```
fisehara, line 512: same applies here,  with the special case that the SubtractDateDate returns an integer on postgres

fisehara (changes requested):
The parentheses logic for date arithmetic is bound to the node not only to the right operand. 
I've added a test cases, that is currently failing but should executed properly.

fisehara (approved):
:+1:

PR #190, opened 2023-02-07
Optimize schema during compilation

Ereski:
This optimizes and normalizes rules before compiling the schema.

Change-type: minor
Signed-off-by: Carol Schulze <carol@balena.io>


Ereski:
@balena-ci I self-certify!

PR #199, opened 2023-03-09
Adds Cast node type to tree parsing

otaviojacobi:
Change-type: patch

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -141,7 +141,7 @@ export type AddDateDurationNode = [
 export type BindNode = ['Bind', number | string] | ['Bind', string, string];
 export type CastNode = ['Cast', AbstractSqlType, string];
 export type CoalesceNode = [
-	'Cast',
+	'Coalesce',
 	UnknownTypeNodes,
 	UnknownTypeNodes,
 	...UnknownTypeNodes[],
--- a/src/referenced-fields.ts
+++ b/src/referenced-fields.ts
@@ -8,12 +8,14 @@ import {
 	AndNode,
 	AnyNode,
 	AverageNode,
+	CastNode,
 	CountNode,
 	CrossJoinNode,
 	DateTruncNode,
 	EngineInstance,
 	EqualsNode,
 	ExistsNode,
+	ExtractJSONPathAsTextNode,
 	FieldsNode,
 	FromNode,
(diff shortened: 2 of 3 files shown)


src/referenced-fields.ts
```diff
@@ -497,6 +497,7 @@ const countTableSelects = (
 
 		// Uninteresting atomic nodes
 		case 'Boolean':
+		case 'Cast':
```
Page-, line 500: This is definitely not safe as `Cast` can definitely include other nodes, eg
```
CAST((SELECT 1 FROM ...) AS BOOLEAN)
```

src/AbstractSQLCompiler.ts
```diff
@@ -141,7 +141,7 @@ export type AddDateDurationNode = [
 export type BindNode = ['Bind', number | string] | ['Bind', string, string];
 export type CastNode = ['Cast', AbstractSqlType, string];
 export type CoalesceNode = [
-	'Cast',
+	'Coalesce',
```
otaviojacobi, line 144: I think this was a small typo on the `CoalesceNode` typing?

src/referenced-fields.ts
```diff
@@ -361,6 +361,7 @@ const countTableSelects = (
 		case 'Alias':
 		case 'Any':
 		case 'Average':
+		case 'Cast':
```
otaviojacobi, line 364: Added Cast as an unary node - this should count in the case of `CAST(SELECT 1 FROM ....) AS BOOLEAN`

src/referenced-fields.ts
```diff
@@ -403,6 +404,7 @@ const countTableSelects = (
 		case 'AddDateNumber':
 		case 'DateTrunc':
 		case 'Equals':
+		case 'ExtractJSONPathAsText':
```
otaviojacobi, line 407: Added ExtraJsonPathAsText as a Binary node - The first argument will always be a JSON and therefore return zero on the recursion, however the second type can be derived from a complex structure

Page-: Fwiw I could fully see the first argument become complex as we improve json support, eg the first argument could be a `SELECT jsonColumn FROM someTable` in future, either way this will catch that quite happily

Page- (approved):

otaviojacobi:
LGTM

otaviojacobi:
lgtm

PR #205, opened 2023-04-11
Support boolean nodes in a ValuesNode

Page-:
This also deprecates the legacy use of direct true/false entries

Change-type: minor


Page-:
LGTM

PR #217, opened 2023-04-17
Revert tightening of `OrderBy` to only field nodes

Page-:
It can in fact be a lot of different things and that is now also reflected in the typing

Change-type: patch


Page-:
LGTM

PR #221, opened 2023-04-26
Add `modelName` to AbstractSqlModel

fisehara:
As abstractlSqlModel is used for a particular named model, the abstractSqlModel should carry this information itself. In particular this is needed for AbstractSqlQuery optimization that needs to be model translation aware.

Change-type: minor


fisehara:
not needed anymore - for now

PR #227, opened 2023-05-15
Wrap composite comparison operations with parenthesis when necessary

thgreasi:
On top of #226 

Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/pinejs.2015/near/358435946

### DIFF
--- a/src/AbstractSQLRules2SQL.ts
+++ b/src/AbstractSQLRules2SQL.ts
@@ -316,8 +316,8 @@ export const getAbstractSqlQuery = (
 const Comparison = (comparison: keyof typeof comparisons): MatchFn => {
 	return (args, indent) => {
 		checkArgs(comparison, args, 2);
-		const a = AnyValue(getAbstractSqlQuery(args, 0), indent);
-		const b = AnyValue(getAbstractSqlQuery(args, 1), indent);
+		const a = precedenceSafeOpValue(comparison, AnyValue, args, 0, indent);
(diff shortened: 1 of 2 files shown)


src/AbstractSQLRules2SQL.ts
```diff
@@ -394,26 +394,32 @@ const mathOperatorNodeTypes = new Set([
...
-	const numericValue = valueMatchFn(operandAbstractSql, indent);
+	const valueExpr = valueMatchFn(operandAbstractSql, indent);
 	const [childNodeType] = operandAbstractSql;
-	if (mathOperatorNodeTypes.has(childNodeType)) {
-		return `(${numericValue})`;
+	if (
+		(mathOperatorNodeTypes.has(parentNodeType) &&
+			mathOperatorNodeTypes.has(childNodeType)) ||
+		// We need parenthesis for chained boolean comparisons, otherwise PostgreSQL complains.
+		(parentNodeType in comparisons && childNodeType in comparisons)
```
thgreasi, line 411: A case that this doesn't emit properly is `SELECT (1 = 0) BETWEEN (1 < 0) AND ((1 > 0))`, in which case it doesn't include parenthesis, which makes the expression differ from the AST.
For that we would need to define the precedence order like in:
See: https://github.com/balena-io-modules/abstract-sql-compiler/compare/184-operator-precedence-light?expand=1#diff-a0b17b1f96003c01184ccb6d784d37e5e78801c6a7ef679c1ff601df04a5d288R389

PR #233, opened 2023-08-09
Update @balena/sbvr-types to 6.0.0

otaviojacobi:
Update @balena/sbvr-types from 5.0.0 to 6.0.0

Change-type: patch


otaviojacobi:
LGTM

PR #237, opened 2023-11-26
Add support for partial unique indexes

thgreasi:
Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/Pine.20Partial.20UNIQUE.20Index.20based.20rules

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -455,6 +455,7 @@ export interface Index {
 	name?: string;
 	/** For rules converted to partial unique indexes this holds the actual rule expression */
 	description?: string;
+	distinctNulls?: boolean;
 	predicate?: BooleanTypeNodes;
 }
 export interface Check {
@@ -879,10 +880,17 @@ $$;`);
 
 		createSqlElements.push(...foreignKeys);
 		for (const index of table.indexes) {
+			let nullsSql = '';
+			if (index.distinctNulls != null) {
+				nullsSql =
+					index.distinctNulls === false
+						? ` NULLS NOT DISTINCT`
+						: ` NULLS DISTINCT`;
+			}
 			// Non-partial indexes are added directly to the CREATE TABLE statement
 			if (index.predicate == null) {
 				createSqlElements.push(
-					index.type + '("' + index.fields.join('", "') + '")',
+					index.type + nullsSql + '("' + index.fields.join('", "') + '")',
 				);
 				continue;
 			}
@@ -890,7 +898,7 @@ $$;`);
 				throw new Error('No name provided for partial index');
 			}
 			const comment = index.description
-				? `-- ${index.description.split(/\r?\n/).join('\n-- ')}\n`
+				? `-- ${index.description.replaceAll(/\r?\n/g, '\n-- ')}\n`
 				: '';
 			const whereSql = compileRule(
 				index.predicate as AbstractSqlQuery,
@@ -900,7 +908,7 @@ $$;`);
 			createIndexes.push(`\
 ${comment}\
 CREATE ${index.type} INDEX IF NOT EXISTS "${index.name}"
-WHERE (${whereSql});
+ON "${table.name}" ("${index.fields.join('", "')}")${nullsSql}
 WHERE (${whereSql});`);
 		}
 
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -449,6 +449,14 @@ export interface Trigger {
 	level: 'ROW' | 'STATEMENT';
 	when: 'BEFORE' | 'AFTER' | 'INSTEAD OF';
 }
+export interface Index {
+	type: string;
+	fields: string[];
+	name?: string;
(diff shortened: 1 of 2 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -448,6 +451,14 @@ export interface Trigger {
 	level: 'ROW' | 'STATEMENT';
 	when: 'BEFORE' | 'AFTER' | 'INSTEAD OF';
 }
+export interface Index {
+	type: string;
+	fields: string[];
+	name?: string;
+	description?: string;
+	nulls?: string;
+	predicate?: BooleanTypeNodes;
```
thgreasi, line 460: I got the name from the PG CREATE INDEX synopsis.
See: https://www.postgresql.org/docs/16/sql-createindex.html

src/AbstractSQLCompiler.ts
```diff
@@ -448,6 +451,14 @@ export interface Trigger {
 	level: 'ROW' | 'STATEMENT';
 	when: 'BEFORE' | 'AFTER' | 'INSTEAD OF';
 }
+export interface Index {
+	type: string;
+	fields: string[];
+	name?: string;
+	description?: string;
+	nulls?: string;
```
Page-, line 459: I think this should probably be a boolean, eg
```suggestion
	distinctNulls?: boolean;
```
or at the very least specific on the strings, eg
```suggestion
	nulls?: 'DISTINCT' | 'NOT DISTINCT';
```

src/AbstractSQLCompiler.ts
```diff
@@ -869,9 +878,30 @@ $$;`);
...
+				createSqlElements.push(
+					index.type + nullsSql + '("' + index.fields.join('", "') + '")',
+				);
+				continue;
+			}
+			if (index.name == null) {
+				throw new Error('No name provided for partial index');
+			}
+			const comment = index.description
+				? `-- ${index.description.split(/\r?\n/).join('\n-- ')}\n`
```
Page-, line 893: I think this could be
```suggestion
				? `-- ${index.description.replaceAll(/\r?\n/g, '\n-- ')}\n`
```
or it could be
```suggestion
				? index.description.replaceAll(/^/gm, '-- ')}\n`
```

src/AbstractSQLSchemaOptimiser.ts
```diff
@@ -32,6 +32,17 @@ const countFroms = (n: AbstractSqlType[]) => {
 	return count;
 };
 
+export const generateRuleHashAcronym = (
```
Page-, line 35: I don't think `acronym` means what you want you want here?

src/AbstractSQLSchemaOptimiser.ts
```diff
@@ -32,6 +32,17 @@ const countFroms = (n: AbstractSqlType[]) => {
...
+	tableName: string,
+	ruleBody: AbstractSqlType,
+) => {
+	const sha = sbvrTypes.SHA.validateSync(
+		`${tableName}$${JSON.stringify(ruleBody)}`,
+	).replace(/^\$sha256\$/, '');
+	// Trim the trigger to a max of 63 characters, reserving at least 32 characters for the hash
+	return `${tableName.slice(0, 30)}$${sha}`.slice(0, 63);
```
Page-, line 43: I think it'd make sense for this to take inspiration from the roughly equivalent code in https://github.com/balena-io-modules/odata-to-abstract-sql/blob/v6.2.4/src/odata-to-abstract-sql.ts#L393-L396 - the things of note to me are:
1. using base36 allows it to still be a string but also fit as much of the hash in as possible in the same number of characters (vs hex/base 16)
2. adjusting the truncation dynamically based upon the actual strings
3. using a named constant for `MAX_ALIAS_LENGTH` rather than a magic number for which the reasoning is not as clear

thgreasi:
It seems that the latest TS fails to compile on master
See: https://github.com/balena-io-modules/abstract-sql-compiler/pull/246

src/AbstractSQLCompiler.ts
```diff
@@ -574,7 +581,10 @@ const dataTypeValidate: EngineInstance['dataTypeValidate'] = async (
 ) => {
 	// In case one of the validation types throws an error.
 	const { dataType, required = false } = field;
-	const validateFn = validateTypes[dataType as keyof typeof sbvrTypes];
+	// Without the `: SbvrType['validate']` widening TS complains that
+	// "none of those signatures are compatible with each other".
+	const validateFn: SbvrType['validate'] =
+		validateTypes[dataType as keyof typeof sbvrTypes];
```
thgreasi, line 587: That's on master atm, and will be dropped from this PR once I rebase.

package.json
```diff
@@ -36,7 +36,7 @@
     "lint-staged": "^13.2.1",
     "mocha": "^10.2.0",
     "ts-node": "^10.9.1",
-    "typescript": "^5.0.4"
+    "typescript": "^5.4.3"
```
thgreasi, line 39: Will be dropped once I rebase on top of master

Page- (approved):

PR #250, opened 2024-06-12
Adds `EndsWithNode` and `ContainsNode` typings

Page-:
Change-type: minor

### DIFF
--- a/src/AbstractSQLOptimiser.ts
+++ b/src/AbstractSQLOptimiser.ts
@@ -531,6 +531,18 @@ const SubtractDateMatcher = tryMatches<
 	),
 );
 
+const EndsWithMatcher = rewriteMatch(
+	'Endswith',
+	[TextValue, TextValue],
+	Helper<MatchFn<LikeNode>>(
+		([haystack, needle]: [TextTypeNodes, TextTypeNodes]) => [
+			'Like',
+			haystack,
+			['Concatenate', ['EmbeddedText', '%'], ['EscapeForLike', needle]],
+		],
+	),
+);
+
 const typeRules = {
 	UnionQuery: (args): UnionQueryNode => {
 		checkMinArgs('UnionQuery', args, 2);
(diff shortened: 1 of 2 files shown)


src/AbstractSQLOptimiser.ts
```diff
@@ -1552,17 +1564,8 @@ const typeRules = {
...
-		Helper<MatchFn<LikeNode>>(
-			([haystack, needle]: [TextTypeNodes, TextTypeNodes]) => [
-				'Like',
-				haystack,
-				['Concatenate', ['EmbeddedText', '%'], ['EscapeForLike', needle]],
-			],
-		),
-	),
+	Endswith: EndsWithMatcher,
+	EndsWith: EndsWithMatcher,
```
otaviojacobi, line 1568: @Page- this is duplicated, I think you want to remove one

src/AbstractSQLOptimiser.ts
```diff
@@ -531,6 +531,18 @@ const SubtractDateMatcher = tryMatches<
 	),
 );
 
+const EndsWithMatcher = rewriteMatch(
```
otaviojacobi, line 534: I don't see why you need to extract this piece if it is not used anywhere else :thinking:

src/AbstractSQLOptimiser.ts
Page-: This is to support both `['Endswith', ...]` and `['EndsWith', ...]`

src/AbstractSQLOptimiser.ts
Page-: It's used in 2 places and I didn't want to copy/paste it in both

src/AbstractSQLOptimiser.ts
otaviojacobi: :facepalm: thanks

otaviojacobi (approved):

PR #262, opened 2025-01-20
Add EqualsAny node

otaviojacobi:
Change-type: minor

### DIFF
--- a/src/AbstractSQLRules2SQL.ts
+++ b/src/AbstractSQLRules2SQL.ts
@@ -82,6 +82,7 @@ const UnknownValue: MetaMatchFn = (args, indent) => {
 		case 'Null':
 		case 'Field':
 		case 'ReferencedField':
+		case 'EqualsAny':
 		case 'Bind':
 		case 'Cast':
 		case 'Coalesce':
@@ -246,6 +247,7 @@ export const isNotNullable = (node: AbstractSqlType): boolean => {
 		case 'Integer':
 		case 'IsDistinctFrom':
 		case 'IsNotDistinctFrom':
+		case 'EqualsAny':
 		case 'Exists':
 		case 'NotExists':
 			return true;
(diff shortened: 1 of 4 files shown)


src/AbstractSQLRules2SQL.ts
```diff
@@ -1238,9 +1238,13 @@ const typeRules: Dictionary<MatchFn> = {
 			throw new SyntaxError('Any not supported on: ' + engine);
 		}
 		const value = AnyValue(getAbstractSqlQuery(args, 0), indent);
-		const innerType =
-			sbvrTypes[args[1] as keyof typeof sbvrTypes].types[engine];
-		return `ANY(CAST(${value} AS ${innerType}[]))`;
+		const castType = args[1] as keyof typeof sbvrTypes;
```
otaviojacobi, line 1241: If https://github.com/balena-io-modules/odata-to-abstract-sql/pull/163/files#r1922431368 is properly addressed this might become unnecessary.

src/AbstractSQLRules2SQL.ts
```diff
@@ -1000,6 +1002,10 @@ const typeRules: Dictionary<MatchFn> = {
 		return `COALESCE(JSON_AGG(${field}), '[]')`;
 	},
 	Equals: Comparison('Equals'),
+	EqualsAny: (args, indent) => {
+		checkArgs('EqualsAny', args, 2);
+		return `${AnyValue(getAbstractSqlQuery(args, 0), indent)} = ANY(${AnyValue(getAbstractSqlQuery(args, 1), indent)})`;
```
Page-, line 1007: Ideally it'd be possible to do 
```suggestion
		return `${AnyValue(getAbstractSqlQuery(args, 0), indent)} = ${typeRules.Any(getAbstractSqlQuery(args, 1), indent)}`;
```
to reuse the same `Any` node for the relevant part, although I'm fine with it as-is and leave that as a potential future thing

otaviojacobi: @Page- They are currently fairly similar, the only reason why can't currently be done this way is because `Any` requires a innerType to be casted, which I am not sure we want to infer for `in`. Ofc I can rework the Any logic to be able to handle no parameter on the innerTyping but at this point we could just use the `AnyNode` anyway. I think if we decide to do it, it should be part of a follow up PR

Page-: Yeah, I just wanted to raise awareness to consider that for future

Page- (approved):

PR #265, opened 2025-02-14
Tests

Page-:


test/abstract-sql/test.ts
```diff
@@ -0,0 +1,192 @@
...
+		| string,
+	expected: string,
+) => void;
+const equals: SqlEquals = (actual, expected) => {
+	if (typeof actual !== 'string') {
+		if (Array.isArray(actual)) {
+			throw new Error('Expected a single query, got multiple');
+		}
+		actual = actual.query;
+	}
```
Page-, line 37: I took a minor liberty in this section so that it's possible to do `sqlEquals(result, '...')` and not have to worry about the `SqlResult | [SqlResult, SqlResult]` typing in every single place when `[SqlResult, SqlResult]` only applies to `UpsertNode`s (very very uncommon case)

test/odata/test.js
```diff
@@ -66,28 +66,25 @@ const sqlEquals = {
...
+			break;
+		case 2:
+			[method, expectation] = args;
+			break;
+		case 3:
+			[method, expectedBindings, expectation] = args;
+			break;
+		case 4:
+			[method, expectedBindings, body, expectation] = args;
+			break;
```
joshbwlng, line 87: Definitely much clearer than the multi-layered conditionals from before. :+1: 
Just in case I also double-checked that the order is correct as it looked a bit odd to me at first (although it obviously is correct as the tests would fail horribly otherwise).

test/abstract-sql/test.ts
joshbwlng: I guess this is never triggered though as the tests pass. Why even allow for `SqlResult[]` if we always throw?

thgreasi (approved):

PR #272, opened 2025-04-07
Switch to a more standard `isFieldTypeNode` helper

Page-:


src/AbstractSQLCompiler.ts
```diff
@@ -666,6 +666,10 @@ export const isSelectNode = (n: AbstractSqlType): n is SelectNode =>
 	n[0] === 'Select';
 export const isWhereNode = (n: AbstractSqlType): n is WhereNode =>
 	n[0] === 'Where';
+export const isFieldTypeNode = (
+	n: AbstractSqlType,
+): n is FieldNode | ReferencedFieldNode =>
+	n[0] === 'Field' || n[0] === 'ReferencedField';
```
otaviojacobi, line 672: @Page- is it prefereable to have the logic duplicated and split like this instead of just having something like

```typescript
export const isFieldTypeNode = (
	n: AbstractSqlType
) => {
	return isFieldValue(n[0]);
}
```

Page-: I've removed the `isFieldValue` in preference of this `isFieldNode` so there isn't any duplication, the only extension I think could make sense is to change it to `isFieldNode(n) || isReferencedFieldNode(n)`

PR #273, opened 2025-04-07
Improve typings for the `In`/`NotIn` optimization helpers

Page-:
Change-type: patch

### DIFF
--- a/src/AbstractSQLOptimiser.ts
+++ b/src/AbstractSQLOptimiser.ts
@@ -107,6 +107,7 @@ import type {
 	StartsWithNode,
 	EscapeForLikeNode,
 	EqualsAnyNode,
+	NotInNode,
 } from './AbstractSQLCompiler';
 import { isFieldTypeNode } from './AbstractSQLCompiler';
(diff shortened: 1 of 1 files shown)


otaviojacobi (approved):

src/AbstractSQLOptimiser.ts
```diff
@@ -1025,13 +1035,13 @@ const typeRules = {
 						return;
 					}
 				} else if (arg[0] === 'NotIn') {
-					const fieldRef = arg[1] as string;
```
thgreasi, line 1028: Was this really a string?
B/c `NotInNode[1]` gives `FieldNode | ReferencedFieldNode` !?!?!?
image

Page-: It was not a string, but it would have been automatically coerced into a string when used to index the object - this is almost certainly from having previously been ometajs and mistyped when converting to typescript

Page-: Or maybe it *was* a string when first typed but then other changes happened, either way it's typed better now

PR #285, opened 2025-06-20
Compile computed fields into the table definition

Page-:
This allows all downstream consumers to simplify as they only ever have to look at/use the table definition and can completely ignore whether fields are computed or not as it will have been handled for them already as necessary

Change-type: minor

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -857,9 +857,12 @@ ${compileRule(definitionAbstractSql as AbstractSqlQuery, engine, true).replace(
 		const createSqlElements: string[] = [];
 		const createIndexes: string[] = [];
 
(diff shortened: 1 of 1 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -857,9 +857,12 @@ ${compileRule(definitionAbstractSql as AbstractSqlQuery, engine, true).replace(
 		const createSqlElements: string[] = [];
 		const createIndexes: string[] = [];
 
+		const computedFields: Array<AbstractSqlField['computed']> = [];
```
joshbwlng, line 860: Could you do something like this to avoid the cast later on?
```typescript
const computedFields: Array<['Alias', AbstractSqlQuery, string]> = [];
```

PR #290, opened 2025-07-09
Postgres: Optimize column order to reduce padding

joshbwlng:
Change-type: patch

### DIFF
--- a/src/AbstractSQLCompiler.ts
+++ b/src/AbstractSQLCompiler.ts
@@ -24,6 +24,7 @@ import {
 	getModifiedFields,
 	insertAffectedIdsBinds,
 } from './referenced-fields';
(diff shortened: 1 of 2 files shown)


src/AbstractSQLCompiler.ts
```diff
@@ -746,6 +747,66 @@ export function compileRule(
...
+	'Big Serial': 1,
+	'Date Time': 1,
+	Integer: 2,
+	Serial: 2,
+	'Short Text': 3,
+	File: 4,
+	Text: 4,
+	// eslint-disable-next-line id-denylist
+	Boolean: 5,
+} as const;
```
joshbwlng, line 764: We could potentially also store sorting weights as a property of the types themselves in `sbvr-types`.

PR #299, opened 2025-08-08
Add the ability to have computed fields generate/call functions

Page-:
https://balena.fibery.io/Work/Project/1766

Change-type: minor

### DIFF
--- a/src/abstract-sql-compiler.ts
+++ b/src/abstract-sql-compiler.ts
@@ -247,6 +247,7 @@ export type CoalesceNode = [
 	...AnyTypeNodes[],
 ];
 export type AnyNode = ['Any', AnyTypeNodes, string];
+export type FnCallNode = ['FnCall', string, ...AnyTypeNodes[]];
 export type UnknownTypeNodes =
 	| SelectQueryNode
 	| UnionQueryNode
@@ -257,7 +258,8 @@ export type UnknownTypeNodes =
 	| CastNode
 	| CaseNode
 	| CoalesceNode
-	| AnyNode;
+	| AnyNode
+	| FnCallNode;
 
 export type ToJSONNode = ['ToJSON', AnyTypeNodes];
 export type AggregateJSONNode = [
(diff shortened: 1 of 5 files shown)


src/abstract-sql-compiler.ts
```diff
@@ -869,10 +882,47 @@ ${compileRule(definitionAbstractSql as AbstractSqlQuery, engine, true).replace(
...
+						throw new Error(
+							`Unknown data type '${dataType}' for engine: ${engine}`,
+						);
+					}
+					if (typeof sqlType === 'function') {
+						sqlType = sqlType.castType;
+					}
+					const fnSignature = `"computed_${table.name}_${fieldName}"("${table.name}" "${table.name}")`;
+					createComputedFunctions.push(`\
+CREATE FUNCTION ${fnSignature}
```
joshbwlng, line 904: Should we `CREATE OR REPLACE FUNCTION` to be safe?

src/abstract-sql-compiler.ts
```diff
@@ -869,10 +882,47 @@ ${compileRule(definitionAbstractSql as AbstractSqlQuery, engine, true).replace(
...
+						sbvrTypes[dataType as keyof typeof sbvrTypes]?.types[engine];
+					if (sqlType == null) {
+						throw new Error(
+							`Unknown data type '${dataType}' for engine: ${engine}`,
+						);
+					}
+					if (typeof sqlType === 'function') {
+						sqlType = sqlType.castType;
+					}
+					const fnSignature = `"computed_${table.name}_${fieldName}"("${table.name}" "${table.name}")`;
```
joshbwlng, line 902: Should we trim function names like we do for triggers with `slice(0, 63)`?

src/abstract-sql-compiler.ts
```diff
@@ -869,10 +882,47 @@ ${compileRule(definitionAbstractSql as AbstractSqlQuery, engine, true).replace(
...
+					createComputedFunctions.push(`\
+CREATE FUNCTION ${fnSignature}
+RETURNS ${sqlType} AS $fn$
+	${compileRule(
```
joshbwlng, line 906: Question about computed fields in general, but should we worry about infinite recursion if a computed field attempts to reference itself directly or even indirectly via a different computed field?

src/abstract-sql-compiler.ts
Page-: Yeah I've trimmed it, along with shrinking the prefix and allowing a name of choice for the function to be specified (in case it becomes necessary, and also avoiding copy/pasting the fn name construction in multiple places)

PR #316, opened 2026-01-22
Optimize conditional uniqueness rules to partial unique indexes

thgreasi:
Change-type: minor
See: https://balena.fibery.io/Work/Project/961


src/schema-optimizations/check-constraint.ts
thgreasi: Extracted mostly as is from src/abstract-sql-schema-optimizer.ts

PR #318, opened 2026-01-30
Add Upper/Lower range functions

jaomaloy:
Allow querying upper/lower limits
of a range such as a date time range
(Tstzrange).

Change-type: patch

### DIFF
--- a/src/abstract-sql-rules-to-sql.ts
+++ b/src/abstract-sql-rules-to-sql.ts
@@ -89,6 +89,8 @@ const UnknownValue: MetaMatchFn = (args, indent) => {
 		case 'TextArray':
 		case 'FnCall':
 		case 'JSONPopulateRecord':
+		case 'RangeLower':
+		case 'RangeUpper':
 			return typeRules[type](rest, indent);
 		case 'SelectQuery':
 		case 'UnionQuery': {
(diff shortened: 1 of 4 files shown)


src/abstract-sql-rules-to-sql.ts
```diff
@@ -197,7 +197,9 @@ export const isDateValue = (type: unknown): type is StrictDateTypeNodes[0] => {
 		type === 'AddDateNumber' ||
 		type === 'AddDateDuration' ||
 		type === 'SubtractDateDuration' ||
-		type === 'SubtractDateNumber'
+		type === 'SubtractDateNumber' ||
+		type === 'RangeLower' ||
+		type === 'RangeUpper'
```
thgreasi: I'm not an expert on this, but since LOWER & UPPER can also return NULL and we could also use then on non-date ranges as well, I think they should be handled directly in AnyValue and not in here.
Probably better to have it in UnknownValue, just like we have COALESCE in there already.

thgreasi (approved):
LGTM but I would like @Page- to also take a look.

src/abstract-sql-rules-to-sql.ts
```diff
@@ -201,6 +203,12 @@ export const isDateValue = (type: unknown): type is StrictDateTypeNodes[0] => {
 	);
 };
 const DateValue = MatchValue(isDateValue);
+// No dedicated range-type nodes exist yet; RangeValue currently falls through to UnknownValue.
+export const isRangeValue = (type: unknown): type is never => {
+	void type;
```
Page-: You should be able to do this via
```suggestion
// eslint-disable-next-line @typescript-eslint/no-unused-vars
export const isRangeValue = (_type: unknown): type is never => {
```

Page- (approved):

thgreasi (approved):

src/abstract-sql-rules-to-sql.ts
```diff
@@ -201,6 +203,11 @@ export const isDateValue = (type: unknown): type is StrictDateTypeNodes[0] => {
 	);
 };
 const DateValue = MatchValue(isDateValue);
+// No dedicated range-type nodes exist yet; RangeValue currently falls through to UnknownValue.
```
thgreasi, line 206:
```suggestion
// No dedicated range-type nodes exist yet; RangeValue currently falls through to UnknownValue.
// eslint-disable-next-line @typescript-eslint/no-unused-vars
```

